PR #17, opened 2024-04-15
Using only paths filter endpoint instead of dedicated ones

Raalsky:


src/neptune_fetcher/cache.py
```diff
@@ -0,0 +1,49 @@
...
+
+class FieldsCache(Dict[str, Union[Field, Series]]):
+    def __init__(self, backend: NeptuneBackend, container_id: QualifiedName, container_type: ContainerType):
+        super().__init__()
+        self._backend: NeptuneBackend = backend
+        self._container_id: QualifiedName = container_id
+        self._container_type = container_type
+        self._field_to_fetchable_visitor = FieldToFetchableVisitor()
+
+    def __cache_miss(self, paths: List[str]) -> None:
```
AleksanderWWW, line 28: Why do we want to use a double underscore here? Name mangling in python afaik is not a recommended practice
https://stackoverflow.com/questions/7456807/should-i-use-name-mangling-in-python

src/neptune_fetcher/cache.py
```diff
@@ -0,0 +1,49 @@
...
+        missed_paths = [path for path in paths if path not in self]
+        if missed_paths:
```
AleksanderWWW, line 30: Could we make it into
```
if not missed_paths:
    return

... <rest of the logic>
```
?
Less indent, a little easier to read

Raalsky: Done

src/neptune_fetcher/cache.py
Raalsky: Done

AleksanderWWW (approved):

PR #20, opened 2024-04-18
Regex support with `columns_regex` and `names_regex` for `fetch_runs_df`

Raalsky:


src/neptune_fetcher/read_only_project.py
```diff
@@ -191,13 +202,56 @@ def fetch_runs_df(
...
+            )
+            regex = re.compile(names_regex)
+            with_ids = []
+
+            for experiment in objects:
+                for field in experiment.fields:
+                    if field.path == "sys/name" and regex.match(field.value) is not None:
+                        with_ids.append(experiment.object_key)
+
+            if with_ids is not None and len(with_ids) > MAX_REGEXABLE_RUNS:
```
AleksanderWWW, line 247: `with_ids` will not be `None`, because it was set to an empty list in line `240`. It's a name collision as we're overwriting what a user gave us as an argument.

src/neptune_fetcher/read_only_project.py
```diff
@@ -191,13 +202,56 @@ def fetch_runs_df(
...
+                        extract_entries=lambda data: data.entries,
+                        project_id=self._project_qualified_name,
+                        field_name_regex=columns_regex,
+                        experiment_ids_filter=with_ids,
+                    )
+                )
+                for field_definition in field_definitions:
+                    columns.add(field_definition.path)
+
+            if columns is not None and len(columns) > MAX_COLUMNS_ALLOWED:
```
AleksanderWWW, line 223: `if columns is not None` is already being checked in line `210`

README.md
```diff
@@ -124,18 +124,20 @@ Fetches the runs' metadata and returns them as a pandas DataFrame.
...
+|-----------------|-----------------------------------------------|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `columns`       | `List[str]`, optional                         | `None`              | Names of columns to include in the table, as a list of field names. The Neptune ID (`"sys/id"`) is included automatically. If `None`, all the columns of the experiments table are included.                                                                                           |
+| `columns_regex` | `str`, optional                               | `None`              | A regex pattern to filter the columns by name in addition to `columns`.                                                                                                                                                                                                                |
```
normandy7, line 130: Just to make this a bit more explicit

```suggestion
| `columns_regex` | `str`, optional                               | `None`              | A regex pattern to filter columns by name. Use this parameter to include columns in addition to the ones specified by the `columns` parameter. |
```

README.md
```diff
@@ -144,10 +146,21 @@ runs_df = project.fetch_runs_df(
 	columns=["sys/name", "sys/modification_time", "training/lr"],
 )
 
+# Fetch all runs with specific columns and matching a regex pattern
```
normandy7, line 149:
```suggestion
# Fetch all runs with specific columns and extra columns that match a regex pattern
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -160,6 +168,9 @@ def fetch_runs_df(
         Args:
             columns: None or a list of column names to include in the result.
                 Defaults to None, which includes all available columns up to 10k.
+            columns_regex: A regex pattern to filter the columns by name in addition to `columns`.
```
normandy7, line 171:
```suggestion
            columns_regex: A regex pattern to filter columns by name.
                Use this parameter to include columns in addition to the ones specified by the `columns` parameter.
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -191,13 +202,56 @@ def fetch_runs_df(
...
+                        experiment_ids_filter=with_ids,
+                    )
+                )
+                for field_definition in field_definitions:
+                    columns.add(field_definition.path)
+
+            if columns is not None and len(columns) > MAX_COLUMNS_ALLOWED:
+                raise ValueError(
+                    f"Too many columns requested ({len(columns)}). "
+                    "Please limit the number of columns to 10k or fewer."
```
normandy7, line 226:
```suggestion
                    "Please limit the number of columns to 10 000 or fewer."
```

normandy7 (commented):
Looks good otherwise

src/neptune_fetcher/read_only_project.py
Raalsky: Thanks, you're right.

src/neptune_fetcher/read_only_project.py
Raalsky: You're right. I've reworked it a little bit.

normandy7 (approved):

AleksanderWWW (approved):

PR #60, opened 2024-09-12
Fetching only up to cumulative length of columns

Raalsky:


kgodlewski (approved):

src/neptune_fetcher/read_only_project.py
```diff
@@ -734,7 +739,12 @@ def filter_columns_regex(
     )
 
     for field_definition in field_definitions:
+        if cumulative_length + len(field_definition.path) > MAX_CUMULATIVE_COLUMN_LENGTH:
+            logger.warning(f"Too many characters in the columns requested. Fetching only {len(columns)} first columns.")
```
normandy7, line 743: > Too many characters in the columns requested.

Does it effectively refer to the combined length of column titles?

> Fetching only {len(columns)} first columns.

What is `len(columns)` in this case?

Raalsky: `len(columns)` -> # of columns

Raalsky: > Does it effectively refer to the combined length of column titles?

Yes, sum of lengths like of column names. Like `['sys/name', 'sys/id']` has a combined length of `14` characters.

normandy7: Sure, but I don't understand the relationship between "requested columns" and this "# of columns"

Raalsky: That's the result of columns regex. So you've provided for instance `config/.*` and you may end up in three situations:
- we've matched `9875` column names and that's above the limit of `5000` different columns -> will be truncated to first `5000`
- we've matched `4321` columns but the cumulative length exceeded the limit of 100k -> will be truncated to as many as possible to fit into the limit for instance `4300`.
- you're under the limits -> good to go, all columns will be included

Raalsky: Yes

normandy7, line 743:
```suggestion
            logger.warning(f"Too many characters in the combined column titles. Fetching only the first {len(columns)} columns.")
```

PR #63, opened 2024-09-17
Rework how data is downloaded in `ReadOnlyProject.fetch*_df()` methods

kgodlewski:
Main changes:

- We now use a new `neptune-client` API call: `query_fields_within_project()` that allows us to directly apply column filters, and returns values in a single call
- The `match_columns_to_filters` now defaults to `True`, as it is obsolete in the current approach: we always fetch columns within a set of runs, not in the context of a project.
- Use protobuf endpoints for downloading values
- E2E tests added, along with a script to populate tests data
- NOTE: some integration tests are skipped in this PR, as their use cases are covered in E2E tests, and fixing the all the integration parts can be done as a next step


src/neptune_fetcher/read_only_project.py
```diff
@@ -260,18 +275,21 @@ def fetch_runs_df(
                 If True: return only trashed runs.
                 If False (default): return only non-trashed runs.
                 If None: return all runs.
-            limit: How many entries to return at most. If `None`, all entries are returned.
+            limit: The maximum number of rows (runs) to return. If `None`, all entries are returned, up to
+                   the hard limit of 50000.
```
normandy7, line 279: ? 

```suggestion
                   the hard limit of 5000.
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -350,17 +367,21 @@ def fetch_experiments_df(
                 If True: return only trashed experiments.
                 If False (default): return only non-trashed experiments.
                 If None: return all experiments.
-            limit: How many entries to return at most. If `None`, all entries are returned.
+            limit: The maximum number of rows (experiments) to return. If `None`, all entries are returned, up to
+                   the hard limit of 50000.
```
normandy7, line 371: ?

```suggestion
                   the hard limit of 5000.
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -379,7 +400,15 @@ def fetch_experiments_df(
             experiments_df = my_project.fetch_experiments_df(query="(accuracy: float > 0.88) AND (loss: float < 0.2)")
             ```
         """
-        return self._fetch_project_objects_df(
+        if names_regex is not None and not names_regex:
+            raise ValueError("The `names_regex` parameter cannot be an empty string.")
```
normandy7, line 404:
```suggestion
            raise ValueError("The `names_regex` argument can't be an empty string.")
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -417,17 +445,27 @@ def _fetch_project_objects_df(
...
+                "or 'tags' arguments."
             )
 
-        prepared_query = _resolve_query(
+        if limit is not None:
+            if limit <= 0:
+                raise ValueError("The 'limit' argument must be greater than 0.")
+            elif limit > MAX_RUNS_ALLOWED:
+                raise ValueError(
+                    f"The 'limit' argument is greater than the maximum allowed value of {MAX_RUNS_ALLOWED}."
```
normandy7, line 460:
```suggestion
                    f"The 'limit' argument can't be higher than {MAX_RUNS_ALLOWED}."
```

README.md
```diff
@@ -286,17 +284,18 @@ __Parameters__:
...
-| `match_columns_to_filters` | `bool`, optional                              | `False`             | If `True`, the columns regex will only match columns that are present in the runs that pass the run filters. The run filters must match up to 5000 entries. If `False`, the columns regex will match all columns in the project.                                                                                                                                           |
+| `match_columns_to_filters` | `bool`, optional                              | `False`             | DEPRECATED: The argument is now assumed to be always `True` and will be removed in a future release.<br/><br/>If `True`, the columns regex will only match columns that are present in the runs that pass the run filters. The run filters must match up to 5000 entries. If `False`, the columns regex will match all columns in the project.                                                                                                                                           |
```
normandy7, line 287:
```suggestion
| `match_columns_to_filters` | `bool`, optional                              | `True`             | DEPRECATED: The argument is assumed to be `True` and will be removed in a future release.<br/><br/>If `True`, the columns regex only matches against fields that are present in the runs that pass the run filters. The run filters must match no more than 5000 entries. If `False`, the columns regex matches against all fields in the project.                                                                                                                                           |
```

src/neptune_fetcher/read_only_project.py
```diff
@@ -236,7 +251,7 @@ def fetch_runs_df(
         ascending: bool = False,
         progress_bar: Union[bool, Optional[ProgressBarType]] = None,
         query: Optional[str] = None,
-        match_columns_to_filters: bool = False,
+        match_columns_to_filters: bool = True,
```
Raalsky, line 254: This has to be explicitly mentioned in CHANGELOG and communicated

src/neptune_fetcher/read_only_project.py
```diff
@@ -379,7 +400,15 @@ def fetch_experiments_df(
...
             ```
         """
-        return self._fetch_project_objects_df(
+        if names_regex is not None and not names_regex:
+            raise ValueError("The `names_regex` parameter cannot be an empty string.")
+
+        if columns is None:
+            columns = []
+
+        columns = list(columns) + ["sys/name"]
```
Raalsky, line 409: What will happen if it was already provided? Set union instead?

src/neptune_fetcher/read_only_project.py
```diff
@@ -811,7 +874,7 @@ def list_objects_from_project(
     ]
 
 
-def verify_string_collection(
+def _verify_string_collection(
```
Raalsky, line 877: We're using `__all__` so no need to worry about underscores

src/neptune_fetcher/read_only_project.py
```diff
@@ -57,31 +63,40 @@
...
+SYS_COLUMNS = ["sys/id", "sys/name", "sys/custom_run_id"]
 
-MAX_CUMULATIVE_COLUMN_LENGTH = 100000
 MAX_CUMULATIVE_LENGTH = 100000
 MAX_QUERY_LENGTH = 250000
 MAX_COLUMNS_ALLOWED = 5000
 MAX_ELEMENTS_ALLOWED = 5000
 MAX_RUNS_ALLOWED = 5000
-NEPTUNE_FETCH_COLUMNS_STEP_SIZE = "NEPTUNE_FETCH_COLUMNS_STEP_SIZE"
+FETCH_COLUMNS_BATCH_SIZE = getenv_int("NEPTUNE_FETCH_COLUMNS_BATCH_SIZE", 10_000)
```
Raalsky, line 92: I'm not a big fan of evaluating env variables as a part of the module instead of evaluation in runtime. It slows down the imports. Should be a part of class initialization instead

src/neptune_fetcher/read_only_project.py
kgodlewski: Yes, `_ensure_default_columns()` does that

src/neptune_fetcher/read_only_project.py
kgodlewski: I agree as a general rule, however what we're doing here is lightweight, and the class creation will follow right after the import 90% of the time, so I think it's fine in that case

src/neptune_fetcher/read_only_project.py
kgodlewski: I'll leave this as a cleanup task for a minor release

README.md
```diff
@@ -205,21 +205,19 @@ __Parameters:__
...
-| `match_columns_to_filters` | `bool`, optional                              | `False`             | If `True`, the columns regex will only match columns that are present in the runs that pass the run filters. The run filters must match up to 5000 entries. If `False`, the columns regex will match all columns in the project.                                                                                               |
+| `match_columns_to_filters` | `bool`, optional                              | `False`             | DEPRECATED: The argument is now assumed to be always `True` and will be removed in a future release.<br/><br/>If `True`, the columns regex will only match columns that are present in the runs that pass the run filters. The run filters must match up to 5000 entries. If `False`, the columns regex will match all columns in the project.                                                                                               |
```
normandy7, line 208: Same as for the other method

```suggestion
| `match_columns_to_filters` | `bool`, optional                              | `True`             | DEPRECATED: The argument is assumed to be `True` and will be removed in a future release.<br/><br/>If `True`, the columns regex only matches against fields that are present in the runs that pass the run filters. The run filters must match no more than 5000 entries. If `False`, the columns regex matches against all fields in the project.                                                                                                                                           |
```

normandy7 (approved):

Raalsky (approved):

PR #68, opened 2024-09-25
`ReadOnlyProject` functions that fetch `DataFrame` are now concurrent.

kgodlewski:
Make functions in `ReadOnlyProject` that return `pandas.DataFrame` fetch data concurrently.
This speeds up data download by 20-70 percent, depending on the data layout.

### DIFF
--- a/src/neptune_fetcher/read_only_project.py
+++ b/src/neptune_fetcher/read_only_project.py
@@ -18,6 +18,7 @@ __all__ = [
 ]
 
 import collections
+import concurrent.futures
 import math
 import os
 from enum import Enum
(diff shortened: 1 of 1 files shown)


src/neptune_fetcher/read_only_project.py
```diff
@@ -688,7 +719,7 @@ def _stream_runs(
         columns=["sys/id"],
         limit=limit,
         sort_by=sort_by,
-        step_size=1000,
+        step_size=FETCH_RUNS_BATCH_SIZE,
```
PatrykGala, line 722: `step_size=FETCH_RUNS_BATCH_SIZE,`?

src/neptune_fetcher/read_only_project.py
```diff
@@ -495,33 +497,62 @@ def _fetch_df(
...
-            if len(all_run_ids) > limit:
-                raise ValueError(
-                    f"The number of runs returned exceeds the limit of {limit}. "
-                    "Please narrow down your query or provide a smaller 'limit' "
-                    "as an argument"
+        # Workers fetching attributes in parallel
+        futures = []
+
+        value_count = 0
+        with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
```
PatrykGala, line 504: Please check the number of connections in http client.

PatrykGala: In near future we will use asyncio

PatrykGala (changes requested):
LGTM

src/neptune_fetcher/read_only_project.py
kgodlewski: It was supposed to be this way, as `step_size` controls the page size retrieved from the backend. For some reason it was hardcoded.

michalsosn (approved):

PatrykGala (approved):

PR #78, opened 2024-10-14
feat: Add step range filter

PatrykGala:


src/neptune_fetcher/cache.py
```diff
@@ -62,20 +62,23 @@ def prefetch_series_values(
         use_threads: bool,
         progress_bar: "ProgressBarType" = None,
         include_inherited: bool = True,
+        step_range=(None, None)
```
michalsosn, line 65: the type is missing. `Tuple[Optional[int], Optional[int]]` ?

kgodlewski (dismissed):

src/neptune_fetcher/cache.py
PatrykGala: fixed

README.md
```diff
@@ -580,15 +581,17 @@ loss = run["loss"].fetch_last()
...
+| `include_timestamp` | `bool`                | `True`       | Whether the fetched data should include the timestamp field.                                                                                                                                        |
+| `include_inherited` | `bool`                | `True`       | If True (default), values inherited from ancestor runs are included. To only fetch values from the current run, set to False.                                                                       |
+| `progress_bar`      | `ProgressBarType`     | `None`       | Set to False to disable the download progress bar, or pass a ProgressBarCallback class to use your own progress bar. If set to None or True, the default tqdm-based progress bar is used.           |
+| `step_range`        | `tuple[float, float]` | (None, None) | - left: left boundary of the range (exclusive). If None, it\'s open on the left. <br> - right: (currently not supported) right boundary of the range (inclusive). If None, it\'s open on the right. |
```
michalsosn, line 594: `| step_range | tuple[float, float] | (None, None) | - left: left boundary of the range (exclusive). If None, the range extends indefinitely on the left. <br> - right: (currently not supported) right boundary of the range (inclusive). If None, the range extends indefinitely on the right. |`

Afaik this description would be more precise, because a range being 'open' actually means that it does not include the value at the boundary.

src/neptune_fetcher/fields.py
```diff
@@ -92,10 +93,11 @@ def fetch_values(
         include_timestamp: bool = True,
         include_inherited: bool = True,
         progress_bar: "ProgressBarType" = None,
+        step_range: Tuple[Union[float, None], Union[float, None]] = (None, None),
     ) -> "DataFrame":
         import pandas as pd
 
-        if self.prefetched_data is None or self.include_inherited != include_inherited:
+        if self.prefetched_data is None or self.include_inherited != include_inherited or self.step_range != step_range:
```
michalsosn, line 100: in theory we could return the prefetched data when self.step_range[0] <= step_range[0] ;o

michalsosn (approved):

src/neptune_fetcher/fields.py
PatrykGala: IMO is is not real use case.

PR #82, opened 2024-10-17
Add tests for `ReadOnlyRun`

kgodlewski:
This is an initial batch of tests that don't rely on prepopulated data, but instead log them during test execution.

In order not to interfere with existing tests that do rely on runs in a project having a specific shape, it is possible to point selected test modules to another existing project. This is already configured in GitHub Actions variables and secrets.


tests/README.md
```diff
@@ -1,10 +1,32 @@
...
+
+## Environment variables
+
+* `NEPTUNE_PROJECT` - project name to use. It is set by the test runner script in
+  GitHub setup.
+* `NEPTUNE_API_TOKEN` - API token to use
+
+Test that populate data during execution could interfere with those that rely on a
+specific data shape. To avoid this, set `NEPTUNE_E2E_PROJECT` to a different project
+that already exists.
```
michalsosn, line 28: so, NEPTUNE_PROJECT - tests on prepopulated with data, NEPTUNE_E2E_PROJECT - tests on a specific data shape 
or is the opposite? I don't think it's unambiguous in this description, nor is it clear from the variable names

tests/e2e/conftest.py
```diff
@@ -50,3 +68,68 @@ def id_to_name():
...
+    def _wrap(self, fn: Callable):
+        def wrapped(*args, **kwargs):
+            result = fn(*args, **kwargs)
+            self.run.wait_for_processing()
+            return result
+
+        wrapped._is_wrapped = True
+        setattr(self.run, fn.__name__, wrapped)
+
+        return wrapped
```
michalsosn, line 99:
```
class SyncRun(Run):
    def log(self, *args, **kwargs):
        result = super().log(*args, **kwargs)
        self.wait_for_processing()
        return result
```
hmm, would this code achieve the same thing and be somewhat simpler?

kgodlewski: You're absolutely right! It requires some modification to the tests, but I'll go with it. I started with the wrapper approach for other methods as well, and ended up only needing log*(), so your solution does the thing and is indeed simpler.

tests/README.md
kgodlewski: I changed both the descriptions and variable names for clarity

michalsosn (approved):

PatrykGala (changes requested):
rebase to main

PatrykGala (approved):

PatrykGala (approved):

PR #87, opened 2024-10-21
Make sure `populate_projects.py` raises any errors from worker processes

kgodlewski:


tests/populate_projects.py
```diff
@@ -68,18 +68,27 @@ def populate_run(run, run_id, tags=None):
...
+            )
 
         for x in range(MM_NUM_RUNS // 2, MM_NUM_RUNS):
-            executor.submit(create_runs, project, x + 1, tags=["tail", "tag2"], experiment_name=f"exp{x+1}")
+            futures.append(
+                executor.submit(create_runs, project, x + 1, tags=["tail", "tag2"], experiment_name=f"exp{x+1}")
+            )
+
+        done, _ = concurrent.futures.wait(futures)
+        for f in done:
```
michalsosn, line 90: it doesn't change much but I guess 
`concurrent.futures.as_completed`
could sometimes end the program faster in case if one of the futures end with an exception while the rest takes time to be processed

michalsosn (approved):

tests/populate_projects.py
kgodlewski: Perhaps, though I tried this with `done, _futures = wait(timeout=...)` in a loop, but that didn't change much. This kind of emulates `as_completed` module the wait timout.

PR #95, opened 2024-10-25
Fetch attributes using Run UUIDs instead of sys/id

kgodlewski:
This change is not user-visible. It's about making sure we're not using a deprecated way to identify specific runs. Instead of using `sys/id`, we're using the Run's UUID.

### DIFF
--- a/src/neptune_fetcher/read_only_project.py
+++ b/src/neptune_fetcher/read_only_project.py
@@ -740,7 +740,7 @@ def _batch_run_ids(
         batch.append(run.object_id)
         if len(batch) == batch_size:
             yield batch
-            batch.clear()
+            batch = []
 
     # We could have a leftover partial batch, smaller than requested max size
     if batch:
--- a/src/neptune_fetcher/read_only_project.py
+++ b/src/neptune_fetcher/read_only_project.py
@@ -41,9 +41,7 @@ from neptune.api.models import (
     LeaderboardEntry,
     NextPage,
 )
-from neptune.api.searching_entries import find_attribute
 from neptune.envs import PROJECT_ENV_NAME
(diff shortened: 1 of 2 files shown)


src/neptune_fetcher/read_only_project.py
```diff
@@ -730,21 +728,19 @@ def _stream_runs(
...
-        run_id = find_attribute(entry=run, path="sys/id")
-        if run_id is None:
-            raise NeptuneException("Experiment id missing in server response")
-
-        batch.append(run_id.value)
+        batch.append(run.object_id)
         if len(batch) == batch_size:
             yield batch
-            batch = []
+            batch.clear()
```
michalsosn, line 743: .clear() and resuing the same [] is dangerous
if the user does not copy/use the yielded list immediately but only save the reference somewhere, you'll modify it's contents

tests/integration/conftest.py
```diff
@@ -201,7 +201,7 @@ def get_float_series_values(*args, **kwargs):
 def make_query_fields_entry(sys_id, custom_run_id):
     return QueryFieldsExperimentResult(
         object_key=sys_id,
-        object_id=custom_run_id,
+        object_id=sys_id,
```
michalsosn, line 204: hmm, is the object_key = object_id true for real data as well?
if not, it'd probably be good to assign different values to them, or you're reducing the test sensitivity and making it blind to the confusion/misuse of object_key/object_id

src/neptune_fetcher/read_only_project.py
kgodlewski: That's correct! Will revert it

michalsosn (approved):

PR #111, opened 2024-11-06
Fix conftest for e2e

kgodlewski:
* Don't require `NEPTUNE_PROJECT` env to be set, if `NEPTUNE_PROJECT` is set in a test module
* Don't create the same run twice
* Don't clutter the console with `logging` errors (see https://github.com/pytest-dev/pytest/issues/5502)


tests/e2e/conftest.py
```diff
@@ -21,6 +22,23 @@
 # around names and values of the metadata.
 
 
+@fixture(scope="session", autouse=True)
```
PatrykGala, line 25: Is it still valid?

kgodlewski: Yes

PatrykGala (approved):

PR #114, opened 2024-11-08
feat: Add NEPTUNE_VERIFY_SSL env variable.

PatrykGala:


src/neptune_fetcher/api/api_client.py
```diff
@@ -70,6 +70,7 @@
 from neptune_fetcher.util import NeptuneException
 
 API_TOKEN_ENV_NAME: Final[str] = "NEPTUNE_API_TOKEN"
+NEPTUNE_VERIFY_SSL: Final[bool] = os.environ.get("NEPTUNE_VERIFY_SSL", "1").lower() in {"1", "true"}
```
kgodlewski, line 73: This needs an update to the docs as well cc @normandy7

kgodlewski: Now that I think of it, a better place for this would be the `neptune-api` repo, as we need consistent behaviour both in `scale` and `fetcher`

normandy7: Is it similar to what we have in client `1.x`? https://docs.neptune.ai/api/environment_variables/#neptune_allow_self_signed_certificate

(I'm trying to get around to the Scale environment variables; they're not documented yet...)

PatrykGala: Yes, it is the same environment variable as in client 1.x.

PR #119, opened 2024-11-15
feat: Use threads

PatrykGala:


src/neptune_fetcher/cache.py
```diff
@@ -86,19 +89,34 @@ def prefetch_series_values(
...
+                result = self._backend.fetch_multiple_series_values(
+                    float_series_paths[batch : batch + batch_size],
+                    include_inherited=include_inherited,
+                    container_id=self._container_id,
+                    step_range=step_range,
+                )
+
+                for path, points in result:
+                    points = list(points)
+                    with lock:
```
kgodlewski, line 110: Wouldn't it be better to acquire the lock before the loop?

src/neptune_fetcher/cache.py
```diff
@@ -86,19 +89,34 @@ def prefetch_series_values(
...
-                progress_bar.update()
+        max_workers = int(os.getenv("NEPTUNE_FETCHER_MAX_WORKERS", 10))
+
+        with tqdm(
+            desc="Fetching metrics", total=len(float_series_paths), unit="metrics"
+        ) as progress_bar, ThreadPoolExecutor(max_workers) as executor:
+            lock = threading.Lock()
+            batch_size = 300
+
+            def fetch(batch):
```
kgodlewski, line 100: This parameter could have a more descriptive name, `start_index`?

src/neptune_fetcher/cache.py
```diff
@@ -86,19 +89,34 @@ def prefetch_series_values(
...
+
+                for path, points in result:
+                    points = list(points)
+                    with lock:
+                        self[path].include_inherited = include_inherited
+                        self[path].step_range = step_range
+                        self[path].prefetched_data = points
+                        progress_bar.update()
+
+            futures = executor.map(fetch, range(0, len(float_series_paths), batch_size))
```
kgodlewski, line 116: What does the `batch_size` argument do here? Looking at the docs, seems it's the value for `timeout`

src/neptune_fetcher/cache.py
```diff
@@ -86,19 +89,34 @@ def prefetch_series_values(
...
+
+        with tqdm(
+            desc="Fetching metrics", total=len(float_series_paths), unit="metrics"
+        ) as progress_bar, ThreadPoolExecutor(max_workers) as executor:
+            lock = threading.Lock()
+            batch_size = 300
+
+            def fetch(batch):
+                result = self._backend.fetch_multiple_series_values(
+                    float_series_paths[batch : batch + batch_size],
```
kgodlewski, line 102: If I'm reading this correctly, we will fetch the same metric multiple times:
```python
fetch(0) -> float_series_paths[0:batch_size]
fetch(1) -> float_series_paths[1:batch_size]
```

kgodlewski (changes requested):

src/neptune_fetcher/cache.py
michalsosn: it's a 3rd argument to range `range(0, len(float_series_paths), batch_size)`
not to executor.map

src/neptune_fetcher/cache.py
PatrykGala:
```
float_series_paths = list(range(10))
for i in list(range(0, len(float_series_paths), batch_size)): print(float_series_paths[i: i+batch_size])
[0, 1, 2]
[3, 4, 5]
[6, 7, 8]
[9]
```

src/neptune_fetcher/cache.py
PatrykGala: is the argument in range function: `range(0, len(float_series_paths), batch_size)`

src/neptune_fetcher/cache.py
PatrykGala: done

src/neptune_fetcher/cache.py
kgodlewski: I think I need to clean my glasses then ;)

src/neptune_fetcher/cache.py
kgodlewski: Yes, with the comment above this now makes sense :)

src/neptune_fetcher/cache.py
kgodlewski: It's necessary, because modifying a dict could trigger internal rehashing / growing mechanisms.

BTW I would add a comment why the lock is being held in the loop rather than outside of it, because it's not super obvious on a first glance.

michalsosn (approved):

PR #120, opened 2024-11-20
Change 'field' to 'attribute' in docs

normandy7:
Change mentions of "field" to "attribute" in docstrings and other documentation.


PatrykGala:
In new neptune-fetcher we will use `attribute`

PR #122, opened 2024-11-20
Rename `field` to `attribute` internally.

kgodlewski:
Mostly classes, some parameters.


PatrykGala (approved):

PatrykGala (changes requested):
Please add backward compatibility for 
    AttributeDefinition,
    AttributePointValue,
    AttributeType,

PatrykGala:
In new api we will use attribute.

PR #135, opened 2024-12-06
fix: add tests checking for cross-package conflicts

michalsosn:


michalsosn:
It's supposed to fail because it detect the incompatibility in the current latest versions of neptune-client and neptune-fetcher

PR #149, opened 2025-01-14
feat(alpha): add attribute filter and find_attributes function

michalsosn:


src/neptune_fetcher/alpha/attribute_filter.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def find_attributes(
+    client: AuthenticatedClient,
+    project_id: str,
+    experiment_ids: list[str],
+    attribute_filter: BaseAttributeFilter,
+    batch_size: int = _DEFAULT_BATCH_SIZE,
+    executor: Optional[Executor] = None,
+) -> list[str]:
```
michalsosn, line 80: since it's an internal method, it could return list[AttributeDefinitionDTO] or equivalent
and it may eventually need to so that we can query for values using experiment ids
but it's easy to change so w/e

src/neptune_fetcher/alpha/attribute_filter.py
```diff
@@ -0,0 +1,234 @@
...
+class BaseAttributeFilter(ABC):
+    def __or__(self, other: "BaseAttributeFilter") -> "AttributeFilterAlternative":
+        return self.any(self, other)
+
+    def any(*filters: "BaseAttributeFilter") -> "AttributeFilterAlternative":
+        return AttributeFilterAlternative(*filters)
+
+
+@dataclass
+class AttributeFilter(BaseAttributeFilter):
```
PatrykGala, line 71: IMO we should expose `AttributeFilter` in the file and move `find_attributes` to new package `internal`(?)

src/neptune_fetcher/alpha/attribute_filter.py
```diff
@@ -0,0 +1,234 @@
...
+        ]
+    )
+    if must_match_regexes is not None:
+        params["attributeNameFilter"]["mustMatchRegexes"] = must_match_regexes
+
+    must_not_match_regexes = _variants_to_list(attribute_filter.name_matches_none)
+    if must_not_match_regexes is not None:
+        params["attributeNameFilter"]["mustNotMatchRegexes"] = must_not_match_regexes
+
+    attribute_types = _variants_to_list(attribute_filter.type_in)
```
PatrykGala, line 155: if None, should use all supported attributes.

src/neptune_fetcher/alpha/attribute_filter.py
```diff
@@ -0,0 +1,234 @@
...
+        if self.type_in is None:
+            self.type_in = ["float", "int", "string", "bool", "datetime", "float_series", "string_set"]
+
+
+@dataclass
+class AttributeFilterAlternative(BaseAttributeFilter):
+    filters: list[AttributeFilter]
+
+
+def find_attributes(
```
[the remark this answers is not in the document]
PatrykGala: Rename to `find_attributes_definitions`

src/neptune_fetcher/alpha/retry.py
PatrykGala, line 1: move to internal package

PatrykGala (changes requested):
.

src/neptune_fetcher/alpha/attribute_filter.py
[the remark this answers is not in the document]
michalsosn: if it's in a package called 'internal', then I guess it can stay without the '_'. 
It's not intended for use outside neptune-fetcher, but it's not private to the attribute_filter.py module

src/neptune_fetcher/alpha/retry.py
michalsosn: ok

src/neptune_fetcher/alpha/attribute_filter.py
michalsosn: hmm, I set the default in the definition of attribute_filter

src/neptune_fetcher/alpha/attribute_filter.py
```diff
@@ -0,0 +1,234 @@
...
+
+from .env import NEPTUNE_FETCHER_MAX_WORKERS
+from .retry import backoff_retry
+
+__ALL__ = (
+    "AttributeFilter",
+    "find_attributes",
+)
+
+_DEFAULT_BATCH_SIZE = 1000
```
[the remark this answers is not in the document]
pawel-ludwikow: Underscore is a bit off.

PatrykGala (approved):

PR #151, opened 2025-01-15
chore: use batching in ReadOnlyProject.fetch_read_only_experiments

PatrykGala:

### DIFF
--- a/src/neptune_fetcher/read_only_project.py
+++ b/src/neptune_fetcher/read_only_project.py
@@ -632,18 +632,15 @@ class ReadOnlyProject:
     ) -> Optional[str]:
         if sys_id is not None:
             query = _make_leaderboard_nql(with_ids=[sys_id], trashed=False)
-
             object_type = "run"
 
         if custom_id is not None:
-
             query = _make_leaderboard_nql(custom_ids=[custom_id], trashed=False)
             object_type = "run"
 
         elif experiment_name is not None:
-
             query = _make_leaderboard_nql(names=[experiment_name], trashed=False)
-            object_type = ("experiment",)
+            object_type = "experiment"
 
         container = list(
             list_objects_from_project(
--- a/src/neptune_fetcher/read_only_project.py
+++ b/src/neptune_fetcher/read_only_project.py
@@ -178,32 +178,26 @@ class ReadOnlyProject:
                 If `False`, individual fields are loaded only when accessed. Default is `True`.
         """
 
-        queries = (
-            [
-                _make_leaderboard_nql(is_run=True, with_ids=with_ids),
-                _make_leaderboard_nql(is_run=True, custom_ids=custom_ids),
-            ]
-            if with_ids or custom_ids
-            else [_make_leaderboard_nql(is_run=True)]
-        )
+        queries = []
+        if with_ids:
+            queries.append(_make_leaderboard_nql(is_run=True, with_ids=with_ids))
 
-        returned_runs = set()
-        for queries in queries:
+        if custom_ids:
+            queries.append(_make_leaderboard_nql(is_run=True, custom_ids=custom_ids))
+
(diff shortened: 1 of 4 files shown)


src/neptune_fetcher/read_only_project.py
```diff
@@ -13,9 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-__all__ = [
-    "ReadOnlyProject",
-]
+__all__ = ["ReadOnlyProject", "list_objects_from_project"]
```
PatrykGala, line 16: todo

src/neptune_fetcher/read_only_project.py
```diff
@@ -177,23 +174,48 @@ def fetch_read_only_runs(
...
             custom_ids: List of custom run ids to fetch.
+            eager_load_fields: Whether to eagerly load the run fields definitions.
+                If `False`, individual fields are loaded only when accessed. Default is `True`.
         """
-        for run_id in with_ids or []:
-            yield ReadOnlyRun(read_only_project=self, with_id=run_id)
 
-        for custom_id in custom_ids or []:
-            yield ReadOnlyRun(read_only_project=self, custom_id=custom_id)
+        queries = (
```
pitercl, line 181: I don't understand this logic. Will this make an API call if both `with_ids` and `custom_ids` are empty/None? Will it always make 2 API calls if either is defined?

If there's some magic in `list_objects_from_project` that prevents is, it's not obvious at a glance.

Overall, please refactor to make obvious that 0, 1 or 2 calls are made depending on how many of `with_ids` and `custom_ids` are non-empty.

src/neptune_fetcher/read_only_project.py
```diff
@@ -202,8 +224,18 @@ def fetch_read_only_experiments(
                 ...
             ```
         """
-        for name in names or []:
-            yield ReadOnlyRun(read_only_project=self, experiment_name=name)
+        query = _make_leaderboard_nql(is_run=False, names=names)
```
pitercl, line 227: Similar problem to `fetch_read_only_runs` - please make obvious a call isn't made in case `names` is empty/None.

src/neptune_fetcher/read_only_project.py
```diff
@@ -595,6 +627,35 @@ def _stream_attributes(
             if remaining <= 0 or not next_page_token:
                 break
 
+    def _fetch_sys_id(
```
pitercl, line 630: I don't understand the motivation for having this method here - it's clearly focused on a single run/experiment - why have this as part of a class dedicated to projects? Furthermore, its behavior is somewhat unexpected when multiple arguments are provided.

src/neptune_fetcher/read_only_project.py
```diff
@@ -177,23 +174,48 @@ def fetch_read_only_runs(
...
+        queries = (
+            [
+                _make_leaderboard_nql(is_run=True, with_ids=with_ids),
+                _make_leaderboard_nql(is_run=True, custom_ids=custom_ids),
+            ]
+            if with_ids or custom_ids
+            else [_make_leaderboard_nql(is_run=True)]
+        )
+
+        returned_runs = set()
```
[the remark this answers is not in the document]
PatrykGala: ok

src/neptune_fetcher/read_only_project.py
PatrykGala: it is private method, I want to avoid exporting `list_objects_from_project` from the file, moving it cause a lot of changes/refactors.

src/neptune_fetcher/read_only_project.py
PatrykGala: fixed,

pitercl (approved):

PR #153, opened 2025-01-20
feat: add cached AuthenticatedClient

pawel-ludwikow:
Add new method to get ApiClient - use `AuthenticatedClientBuilder.build(ctx=None)` to fetch cached, context aware client.


src/neptune_fetcher/alpha/fields.py
```diff
@@ -0,0 +1,160 @@
...
+
+if TYPE_CHECKING:
+    from pandas import DataFrame
+
+    from neptune_fetcher.alpha.api_client import ApiClient
+
+T = TypeVar("T")
+
+
+class FieldType(Enum):
```
michalsosn, line 55: hmm, all these Fields and methods for api calls probably shouldn't need to be copied as part of this task
I think we only need to cache neptune_api.client.AuthenticatedClient ?

pawel-ludwikow: File was removed as not needed

src/neptune_fetcher/alpha/api_client.py
```diff
@@ -0,0 +1,63 @@
...
+    get_config_and_token_urls,
+)
+from .context import Context
+from .internal.context import get_context as internal_get_context
+
+# Disable httpx logging, httpx logs requests at INFO level
+logging.getLogger("httpx").setLevel(logging.WARN)
+
+
+class BuildAuthenticatedClient:
```
PatrykGala, line 40: `AuthenticatedClientBuilder` ?

src/neptune_fetcher/alpha/data.py
PatrykGala, line 1: contexy.py?

src/neptune_fetcher/alpha/internal/context.py
```diff
@@ -0,0 +1,44 @@
...
+
+def get_context(ctx: Optional[Context] = None) -> Context:
+    return sanitize_context_falling_back_to_env(ctx or _neptune_context.get())
+
+
+def set_context(ctx: Context):
+    _neptune_context.set(ctx)
+
+
+def sanitize_context_falling_back_to_env(ctx: Context) -> Context:
```
PatrykGala, line 37:
```suggestion
def _sanitize_context_falling_back_to_env(ctx: Context) -> Context:
```

PatrykGala (changes requested):
LGTM,

src/neptune_fetcher/alpha/internal/context.py
```diff
@@ -0,0 +1,44 @@
...
+from ..data import (
+    NEPTUNE_API_TOKEN,
+    NEPTUNE_PROJECT,
+    Context,
+)
+
+_neptune_context = contextvars.ContextVar("neptune_context")
+
+
+def get_context(ctx: Optional[Context] = None) -> Context:
```
PatrykGala, line 29: set and get_context should be a public api, please move to alpha package

pawel-ludwikow: Those are private internals of the implementation of global context handling.

I can rename those methods to be visually different than public ones.

src/neptune_fetcher/alpha/internal/context.py
pawel-ludwikow: I'm reluctant - we're already in the `internal` package. So no one ought to peek here...

src/neptune_fetcher/alpha/api_client.py
pawel-ludwikow: changed

src/neptune_fetcher/alpha/internal/context.py
pawel-ludwikow: Renamed

src/neptune_fetcher/alpha/data.py
pawel-ludwikow: Renamed to something in-between: context_data.py

PatrykGala (approved):

PR #155, opened 2025-01-20
Make sure we don't crash on unknown types returned by the backend.

kgodlewski:
This PR makes `neptune-fetcher` more resilient to attributes of unknown types returned by the backend. The change makes fetcher accept unknown attribute types:
 * emit a warning when an unknown type is received
 * return `None` / empty `DataFrame` on access to the given attribute

The unit tests suite attempts to test every method for the above behaviour, and also for cases when no unknown attributes are returned -> no warnings should be emitted.

### DIFF
--- a/src/neptune_fetcher/util.py
+++ b/src/neptune_fetcher/util.py
@@ -58,3 +58,12 @@ def escape_nql_criterion(criterion):
     """
 
     return criterion.replace("\\", r"\\").replace('"', r"\"")
+
+
+def warn_unsupported_value_type(type_: str) -> None:
+    warnings.warn(
+        f"A value of type `{type_}` was returned by your query. This type is not supported by your installation "
+        "of neptune-fetcher and values will evaluate to `None` and empty DataFrames. "
+        "Upgrade neptune-fetcher to access this data.",
+        NeptuneWarning,
+    )
--- a/src/neptune_fetcher/util.py
+++ b/src/neptune_fetcher/util.py
@@ -60,7 +60,19 @@ def escape_nql_criterion(criterion):
     return criterion.replace("\\", r"\\").replace('"', r"\"")
 
 
+# We keep a set of types we've warned the user about to make sure we warn about a type only once.
+# This is necessary because of a bug in pandas, that causes duplicate warnings to be issued everytime after an
+# DataFrame() is created (presumably only empty DF).
+# The bug basically makes `warnings.simplefilter("once", NeptuneWarning)` not work as expected, and would flood
+# the user with warnings in some cases.
+_warned_types = set()
+
+
 def warn_unsupported_value_type(type_: str) -> None:
+    if type_ in _warned_types:
+        return
+
+    _warned_types.add(type_)
     warnings.warn(
         f"A value of type `{type_}` was returned by your query. This type is not supported by your installed version "
         "of neptune-fetcher. Values will evaluate to `None` and empty DataFrames. "
--- a/src/neptune_fetcher/util.py
+++ b/src/neptune_fetcher/util.py
@@ -62,8 +62,8 @@ def escape_nql_criterion(criterion):
 
 def warn_unsupported_value_type(type_: str) -> None:
     warnings.warn(
-        f"A value of type `{type_}` was returned by your query. This type is not supported by your installation "
-        "of neptune-fetcher and values will evaluate to `None` and empty DataFrames. "
+        f"A value of type `{type_}` was returned by your query. This type is not supported by your installed version "
+        "of neptune-fetcher. Values will evaluate to `None` and empty DataFrames. "
         "Upgrade neptune-fetcher to access this data.",
         NeptuneWarning,
     )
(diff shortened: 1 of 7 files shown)


src/neptune_fetcher/util.py
```diff
@@ -58,3 +58,12 @@ def escape_nql_criterion(criterion):
     """
 
     return criterion.replace("\\", r"\\").replace('"', r"\"")
+
+
+def warn_unsupported_value_type(type_: str) -> None:
+    warnings.warn(
+        f"A value of type `{type_}` was returned by your query. This type is not supported by your installation "
```
PatrykGala, line 65: `installation` -> `version`?

tests/unit/test_unsupported_attributes.py
```diff
@@ -0,0 +1,428 @@
...
+
+def _assert_warning(record, *attr_types):
+    """Assert that a warning about unsupported types, for a given type, was issued exactly once."""
+    if not attr_types:
+        # Default to the two unsupported types used in TestApiClient
+        attr_types = [type_ for _, type_ in DEFAULT_UNSUPPORTED_ATTRS]
+
+    for rec, attr_type in zip(record, attr_types):
+        assert f"of type `{attr_type}`" in rec.message.args[0]
+        assert "not supported" in rec.message.args[0]
```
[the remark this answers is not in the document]
kgodlewski: Good catch! TBH I did that in the first place, but there are cases where a warning is emitted twice for a type, and decided that it's something we can live with, as there are many variations, and ROI of making this super-strict is not really worth it. Might be wrong here. cc @pitercl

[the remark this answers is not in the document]
kgodlewski: I incorrectly read the pytest docs, the checks are now tighter

pitercl (changes requested):

src/neptune_fetcher/util.py
normandy7:
```suggestion
        f"A value of type `{type_}` was returned by your query. This type is not supported by your installed version "
```

src/neptune_fetcher/util.py
```diff
@@ -58,3 +58,12 @@ def escape_nql_criterion(criterion):
...
+        "of neptune-fetcher and values will evaluate to `None` and empty DataFrames. "
```
normandy7, line 66:
```suggestion
        "of neptune-fetcher. Values will evaluate to `None` and empty DataFrames. "
```

tests/unit/test_unsupported_attributes.py
```diff
@@ -0,0 +1,660 @@
...
+    A simple wrapper around pytest.warns, that makes sure we clear the set of warned types after the test.
+    If we didn't do that, we'd get false positives in tests that expect a warning to NOT be emitted.
+
+    See comments for util.py:_warned_types for more details.
+    """
+
+    with pytest.warns(NeptuneWarning, *args, **kwargs) as record:
+        yield record
+
+    neptune_fetcher.util._warned_types.clear()
```
pitercl, line 284:
```suggestion
    neptune_fetcher.util._warned_types.clear()
    with pytest.warns(NeptuneWarning, *args, **kwargs) as record:
        yield record

    neptune_fetcher.util._warned_types.clear()

```

Without this change `test_warn_and_forget_type` is consistently failing in my local env because it runs after `test_warn_unsupported_value_type`. 
I guess we could skip the second `clear()` call, but I think it's a good addition.

The failure is this:
```FAILED         [ 10%]
tests/unit/test_unsupported_attributes.py:320 (test_warn_and_forget_type)
def test_warn_and_forget_type():
>       with warns_and_forgets_types() as record:

test_unsupported_attributes.py:322: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.10/contextlib.py:142: in __exit__
    next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (), kwargs = {}, record = WarningsChecker(record=True)

    @contextmanager
    def warns_and_forgets_types(*args, **kwargs):
        """
        A simple wrapper around pytest.warns, that makes sure we clear the set of warned types after the test.
        If we didn't do that, we'd get false positives in tests that expect a warning to NOT be emitted.
    
        See comments for util.py:_warned_types for more details.
        """
    
>       with pytest.warns(NeptuneWarning, *args, **kwargs) as record:
E       Failed: DID NOT WARN. No warnings of type (<class 'neptune_fetcher.util.NeptuneWarning'>,) were emitted.
E        Emitted warnings: [].

test_unsupported_attributes.py:281: Failed```

kgodlewski: The 2nd clear (after yield) is the one that is crucial. I agree that starting with a clear slate is what we want in all the tests in the end (though initially I did not add the clear on purpose, but we don't have scenarios where it's breaking tests)

PR #160, opened 2025-01-23
chore: add tests for list_experiments

pawel-ludwikow:
Add e2e tests for `list_experiments`


tests/e2e/alpha/test_experiments.py
```diff
@@ -0,0 +1,125 @@
+import os
```
pitercl, line 1: Please add a test for local / global context with different projects.

pawel-ludwikow: I've added another project and tests for global context (local was there already)

PR #166, opened 2025-01-27
fix: colocate e2e tests using same module-scoped experiment fixture

michalsosn:
module-scoped fixture creating a text experiment and writing data to it is recreated on each pytest-xdist worker
I want to group tests that use the same fixture so that it's initialized only once

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,0 +1,5 @@
+def pytest_set_filtered_exceptions() -> list[type[BaseException]]:
+    class DoNotFilterAnythingMarker(Exception):
+        pass
+
+    return [DoNotFilterAnythingMarker]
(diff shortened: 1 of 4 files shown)


tests/unit/conftest.py
```diff
@@ -0,0 +1,5 @@
+def pytest_set_filtered_exceptions() -> list[type[BaseException]]:
+    class DoNotFilterAnythingMarker(Exception):
```
PatrykGala, line 2: Why do we need the filter?

michalsosn: I have enabled pytest-retry (with 3 retries) globally, but don't want it to do anything in unit tests, so I'm configuring it here to only retry when a unit test throws 'DoNotFilterAnythingMarker'.

PatrykGala:
I've removed xdist from pytest.

PR #168, opened 2025-01-28
Make the API client cache thread safe, and refactor it a bit

kgodlewski:
Changes:
 * get rid of `utilities` in favour of a single, already existing `util.py` module
 * add locking to cache operations
 * refactor the class into a function
 * add tests


src/neptune_fetcher/alpha/internal/api_client.py
```diff
@@ -0,0 +1,76 @@
...
+
+_cache: dict[int, AuthenticatedClient] = {}
+_lock: threading.RLock = threading.RLock()
+
+
+def get_client(context: Optional[Context] = None, proxies: Optional[Dict[str, str]] = None) -> AuthenticatedClient:
+    api_token = get_context(context).api_token
+    hash_key = hash((api_token, _dict_to_hashable(proxies)))
+
+    with _lock:
```
pitercl, line 49: As discussed, let's go with a single lock and creating the client within it + `_cache[] = client` at the end.

Plus, a short comment explaining we're aware it can be optimized (lock per hash_key), but for now we're going with simplicity.

tests/unit/alpha/test_api_client_cache.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@fixture
+def default_ctx():
+    return Context(api_token=make_token(), project="default_project")
+
+
+@fixture(autouse=True)
+def set_default_context(monkeypatch, default_ctx):
+    monkeypatch.setenv(env.NEPTUNE_PROJECT.name, default_ctx.project)
```
pitercl, line 64: Why via envs and not just `npt.set_context(default_ctx)`?

tests/unit/alpha/test_api_client_cache.py
```diff
@@ -0,0 +1,142 @@
...
+    clear_cache,
+    get_client,
+)
+
+# Caching logic being tested: only API token and proxies are used as part of the cache key. Projects don't matter.
+
+
+# This hook is called by pytest on test collection. We force all tests in this file to
+# be run in a single process, so that we can test the cache properly.
+def pytest_collection_modifyitems(config, items):
```
pitercl, line 38: Why does this matter? Ech case is self-contained and cleans up after itself.

src/neptune_fetcher/alpha/internal/api_client.py
```diff
@@ -0,0 +1,76 @@
...
+)
+
+# Disable httpx logging, httpx logs requests at INFO level
+logging.getLogger("httpx").setLevel(logging.WARN)
+
+_cache: dict[int, AuthenticatedClient] = {}
+_lock: threading.RLock = threading.RLock()
+
+
+def get_client(context: Optional[Context] = None, proxies: Optional[Dict[str, str]] = None) -> AuthenticatedClient:
```
pitercl, line 45: I think `context` should not be optional.
As per our discussion in another MR, we'll always do `validate_context(context or get_context())` at the beginning of each method - why would we need to do it here too, then? And it will simplify tests - right now they need to care about setting envs, default global context, etc - it's not this module's responsibility.

pitercl (changes requested):

tests/unit/alpha/test_api_client_cache.py
kgodlewski: In case we have some lingering data in the cache, for whatever reason - it's better to reduce the chance of errors slipping by because of separate processes. I updated the comment to reflect that better.

tests/unit/alpha/test_api_client_cache.py
kgodlewski: Right, fixed

pitercl (approved):

PR #171, opened 2025-01-28
chore: Use Generator in fetch attributes definitions.

PatrykGala:

- [x] Did you **update the CHANGELOG**? (not for test updates, internal changes/refactors or CI/CD setup)
- [x] Did you **ask the docs owner** to review all the user-facing changes?

### DIFF
--- a/src/neptune_fetcher/alpha/internal/attribute.py
+++ b/src/neptune_fetcher/alpha/internal/attribute.py
@@ -64,22 +64,7 @@ def fetch_attribute_definitions(
     experiment_identifiers: Iterable[identifiers.ExperimentIdentifier],
     attribute_filter: filter.BaseAttributeFilter,
     batch_size: int = _DEFAULT_BATCH_SIZE,
-) -> list[AttributeDefinition]:
-    items = [
-        page.items
-        for page in stream_attribute_definitions(
-            client, project_identifiers, experiment_identifiers, attribute_filter, batch_size
-        )
-    ]
-    return list(it.chain.from_iterable(items))
-
-
-def stream_attribute_definitions(
-    client: AuthenticatedClient,
-    project_identifiers: Iterable[identifiers.ProjectIdentifier],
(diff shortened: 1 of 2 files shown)


src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -60,43 +64,66 @@ def fetch_attribute_definitions(
     experiment_identifiers: Iterable[identifiers.ExperimentIdentifier],
     attribute_filter: filter.BaseAttributeFilter,
     batch_size: int = _DEFAULT_BATCH_SIZE,
-    executor: Optional[Executor] = None,
 ) -> list[AttributeDefinition]:
```
michalsosn, line 67: I'd just replace fetch_attribute_definitions, we don't need it and you can always just list(..) the generator

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -60,43 +64,66 @@ def fetch_attribute_definitions(
...
+        _attribute_filter: filter.BaseAttributeFilter,
+    ) -> List[filter.AttributeFilter]:
+        if isinstance(_attribute_filter, filter.AttributeFilter):
+            return [_attribute_filter]
+        elif isinstance(_attribute_filter, filter._AttributeFilterAlternative):
+            return list(it.chain.from_iterable(split_to_tasks(child) for child in _attribute_filter.filters))
+        else:
+            raise ValueError(f"Unexpected filter type: {type(_attribute_filter)}")
+
+    with _create_executor() as executor:
```
michalsosn, line 94: I wouldn't remove the executor parameter, but retain the old behavior ie reuse the given executor if defined or create a new one if it's not

Otherwise https://github.com/neptune-ai/neptune-fetcher/blob/4927dfe7602487e1bda1a762e6b445e142fa13bf/src/neptune_fetcher/alpha/internal/metric.py#L128 a code like this may create a lot of executors in parallel

michalsosn (approved):

PR #173, opened 2025-01-29
feat(alpha): fetch experiment table

michalsosn:


src/neptune_fetcher/alpha/experiments.py
```diff
@@ -0,0 +1,127 @@
...
+                for attribute_definition_page in attribute_definition_pages:
+                    attribute_values_pages = _attribute.fetch_attribute_values(
+                        client=client,
+                        project_identifier=project,
+                        experiment_identifiers=[experiment_identifier],
+                        attribute_definitions=attribute_definition_page.items,
+                        executor=executor,
+                    )
+                    for attribute_values_page in attribute_values_pages:
+                        result[experiment.sys_name].extend(attribute_values_page.items)
```
michalsosn, line 122: this way of combining generators gives correct results... but only effectively uses 3 threads
I need to rewrite it to something akin to what Patryk did

src/neptune_fetcher/alpha/experiment.py
```diff
@@ -0,0 +1,126 @@
...
+                experiment_identifier = _identifiers.ExperimentIdentifier(project, experiment.sys_id)
+                attribute_definition_pages = _attribute.fetch_attribute_definitions(
+                    client=client,
+                    project_identifiers=[project],
+                    experiment_identifiers=[experiment_identifier],
+                    attribute_filter=attributes_filter,
+                    executor=executor,
+                )
+                for attribute_definition_page in attribute_definition_pages:
+                    attribute_values_pages = _attribute.fetch_attribute_values(
```
pitercl, line 113: This cannot be called one run at a time -> we absolutely need batching here.

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -240,3 +248,98 @@ def _union_options(options: list[Optional[list[str]]]) -> Optional[list[str]]:
...
+
+
+def fetch_attribute_values(
+    client: AuthenticatedClient,
+    project_identifier: identifiers.ProjectIdentifier,
+    experiment_identifiers: Iterable[identifiers.ExperimentIdentifier],
+    attribute_definitions: Iterable[AttributeDefinition],
+    batch_size: int = env.NEPTUNE_FETCHER_BATCH_SIZE.get(),
+    executor: Optional[Executor] = None,
+) -> Generator[util.Page[AttributeValue], None, None]:
```
pitercl, line 260: The return type must be able to express attributes for multiple experiments

src/neptune_fetcher/alpha/internal/env.py
```diff
@@ -62,3 +62,4 @@ def _map_bool(value: str) -> bool:
 NEPTUNE_PROJECT = EnvVariable[str]("NEPTUNE_PROJECT", _map_str)
 NEPTUNE_VERIFY_SSL = EnvVariable[bool]("NEPTUNE_VERIFY_SSL", _map_bool, True)
 NEPTUNE_FETCHER_MAX_WORKERS = EnvVariable[int]("NEPTUNE_FETCHER_MAX_WORKERS", int, 10)
+NEPTUNE_FETCHER_BATCH_SIZE = EnvVariable[int]("NEPTUNE_FETCHER_BATCH_SIZE", int, 10_000)
```
pitercl, line 65: I think we might need more granular control. If some endpoint doesn't support >10k batch this is effectively dead code - ppl won't be able to increase the batch size at all.

src/neptune_fetcher/alpha/filter.py
```diff
@@ -42,14 +42,12 @@ def any(*filters: "BaseAttributeFilter") -> "BaseAttributeFilter":
...
-    ] = field(
-        default_factory=lambda: list(types.ALL_TYPES)
-    )  # type: ignore
+    type_in: list[Literal["float", "int", "string", "bool", "datetime", "float_series", "string_set"]] = field(
+        default_factory=lambda: list(types.ALL_TYPES)  # type: ignore
+    )
     name_matches_all: Union[str, list[str], None] = None
     name_matches_none: Union[str, list[str], None] = None
-    aggregations: Optional[list[Literal["last", "min", "max", "average", "variance", "auto"]]] = None
+    aggregations: list[Literal["last", "min", "max", "average", "variance"]] = field(default_factory=lambda: ["last"])
```
pitercl, line 50: What happened here? `None` is a much better value for atoms that effectively don't have aggregations, than `last`. 

If reverting this requires some heavy rewrites - pls let me know. Otherwise - pls revert.

src/neptune_fetcher/alpha/internal/types.py
```diff
@@ -12,21 +12,123 @@
...
+# This is necessary because of a bug in pandas, that causes duplicate warnings to be issued everytime after an
+# DataFrame() is created (presumably only empty DF).
+# The bug basically makes `warnings.simplefilter("once", NeptuneWarning)` not work as expected, and would flood
+# the user with warnings in some cases.
+_warned_types = set()
 
-def map_attribute_type_user_to_backend(_type: str) -> str:
-    return _ATTRIBUTE_TYPE_USER_TO_BACKEND_MAP.get(_type, _type)
 
+def warn_unsupported_value_type(type_: str) -> None:
```
pitercl, line 124: This is copy paste, right? I realize reaching to the non-alpha util module is not pretty, but let's at least shortly discuss it. Leaving this comment so that it doesn't get lost.

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -49,21 +52,29 @@
...
 
 def fetch_attribute_definitions(
     client: AuthenticatedClient,
     project_identifiers: Iterable[identifiers.ProjectIdentifier],
     experiment_identifiers: Iterable[identifiers.ExperimentIdentifier],
     attribute_filter: filter.BaseAttributeFilter,
-    batch_size: int = _DEFAULT_BATCH_SIZE,
+    batch_size: int = env.NEPTUNE_FETCHER_BATCH_SIZE.get(),
     executor: Optional[Executor] = None,
 ) -> Generator[util.Page[AttributeDefinition], None, None]:
```
pitercl, line 79: Generator doesn't make sense here. Multiple attribute filters can select one attribute in which case it will get duplicated. They need to be deduplicated before further processing and this deduplication needs some non-trivial logic when it comes to aggregations. 

I propose the following scheme when merging results from FilterA and FilterB:
1. If an attribute was returned by just one filter -> include the attribute in the result
2. If an attribute was returned by both filters and it has the same, non-floatSeries type in both filters -> include the attribute in the result.
3. If an attribute was returned by both filters and it's a float series and (1) both filters specified aggregations explicitly -> include all specified aggregations in the result, (2) only one filter specified an aggregations -> include only the specified aggregations in the result, (3) neither of the filters specified an aggregation -> include "last" in the result.

This is just a proposal, I'm not thinking 100% straight right now, maybe there are some problems with it, maybe something more intuitive can be devised. Let's discuss tomorrow.

pitercl (changes requested):
I'm not finished yet, but wanted to share this partial review anyway.
I haven't seen the tests yet + I need to dig deeper into the logic.

src/neptune_fetcher/alpha/internal/util.py
```diff
@@ -137,3 +141,8 @@ def backoff_retry(
         raise NeptuneException("Unknown error occurred when requesting data")
 
     raise NeptuneException(f"Failed to get response after {tries} retries. " + "\n".join(msg))
+
+
+def create_executor() -> Executor:
```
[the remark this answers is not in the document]
michalsosn: sure

src/neptune_fetcher/alpha/experiment.py
michalsosn: absolutely, changed

src/neptune_fetcher/alpha/internal/attribute.py
michalsosn: in the updated version AttributeValue has a reference to ExperimentIdentifier
there is also an option to return dict[ExperimentIdentifier, list[AttributeValue]], but I'd need to rework util.Page and a reference is probably fine

src/neptune_fetcher/alpha/filter.py
michalsosn: Hmm, it's easy to change. 
I guess I wanted to have the default here, since it's the first place in the program where this information occurs and it's usually good to have the default for a parameter defined at the top level rather than somewhere deep into internals

I interpreted this class differently I think. It's an AttributeFilter, not AttributeDefinition or AttributeValue, so it's neither an atom or a series, but still a query.
The aggregations field says "when you encounter a float_series, project it to a set of these aggregates".
So it's not really that absurd to have it set, because it's unknown what attributes this query will materialize into

src/neptune_fetcher/alpha/internal/types.py
michalsosn: it's from https://github.com/neptune-ai/neptune-fetcher/pull/155/files 
We talked at one meeting that I really should remember to handle this case of an unknown type in the new method

Hmm, I'm trying to write this code completely independent of the code outside of alpha. So that we may just delete everything outside alpha and it won't have any effect on the new fetcher at all. 
So yeah, it's copied, just as backoff_retry or e.g. metrics._fetch_multiple_series_values

src/neptune_fetcher/alpha/internal/attribute.py
michalsosn: and this is the part we discussed on the last meeting
returing Filters with AttributeDefinitions is to be changed

src/neptune_fetcher/alpha/internal/env.py
michalsosn: sure

pitercl (approved):

PR #177, opened 2025-01-29
Add `list_experiments()`

kgodlewski:
The test for `ExperimentFilter` argument is not super extensive, we can add more cases later when we decide on which level we want to achieve full coverage of e2e cases.

Merging into `pg/fetch-metrics` because of the test utilities that exist in this branch.


tests/e2e/alpha/test_experiments.py
```diff
@@ -206,3 +228,84 @@ def test__fetch_metrics_unique(
...
+    to predict the exact output because of the filtering applied"""
+    names = list_experiments(regex)
+    assert len(names) == len(expected)
+    assert set(names) == set(expected)
+
+
+@pytest.mark.parametrize(
+    "filter_, expected",
+    [
+        (ExperimentFilter.eq(Attribute(f"{PATH}/test/int-value", type="int"), 12345), []),
```
pitercl, line 273: If we're checking this (which is good), we should also check the positive scenario - that filtering by int (and other types) actually returns something.

tests/e2e/alpha/test_experiments.py
```diff
@@ -206,3 +228,84 @@ def test__fetch_metrics_unique(
...
+)
+def test_list_experiments_with_regex_matching_some(regex, expected):
+    """This check is more strict than test_list_experiments_with_regex_matching_all, as we are able
+    to predict the exact output because of the filtering applied"""
+    names = list_experiments(regex)
+    assert len(names) == len(expected)
+    assert set(names) == set(expected)
+
+
+@pytest.mark.parametrize(
```
pitercl, line 270: Maybe at least one alternative and negation test?

pitercl (changes requested):

src/neptune_fetcher/alpha/experiment.py
```diff
@@ -203,3 +206,28 @@ def _map_keys_preserving_order(
...
+
+    validated_context = _context.validate_context(context or _context.get_context())
+    client = _api_client.get_client(validated_context)
+
+    if isinstance(experiments, str):
+        experiments = ExperimentFilter.matches_all(Attribute("sys/name", type="string"), regex=experiments)
+
+    assert validated_context.project is not None  # mypy
+    pages = _experiment.fetch_experiment_sys_attrs(
+        client, _identifiers.ProjectIdentifier(validated_context.project), experiments
```
michalsosn, line 231: fetch_experiment_sys_attrs accepts sort_by, sort_direction, limit now that the table PR was merged

pitercl (approved):

PR #182, opened 2025-01-31
Add `list_attributes()`

kgodlewski:
Tests to be added


src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -396,3 +398,45 @@ def _make_new_attributes_page_params(
...
+            batch_size,
+            executor,
+        ):
+            yield from (attr.name for attr in attrs_page.items)
+
+    if attribute_filter is None:
+        attribute_filter = filter.AttributeFilter()
+
+    executor = executor or util.create_thread_pool_executor()
+    with executor:
```
PatrykGala, line 429: context manager closes executor from function argument

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -396,3 +398,45 @@ def _make_new_attributes_page_params(
...
+        matching_experiments = experiment.fetch_experiment_sys_attrs(
+            client,
+            project_identifier=project_id,
+            experiment_filter=experiment_filter,
+            executor=executor,
+        )
+
+        futures = [executor.submit(list_single_page, page) for page in matching_experiments]
```
PatrykGala, line 437: The main thread generates experiments and waits at this line until all of them are complete

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -396,3 +398,45 @@ def _make_new_attributes_page_params(
...
+        ]
+        for attrs_page in fetch_attribute_definitions(
+            client,
+            [project_id],
+            experiment_identifiers,
+            cast(filter.AttributeFilter, attribute_filter),
+            batch_size,
+            executor,
+        ):
+            yield from (attr.name for attr in attrs_page.items)
```
PatrykGala, line 423: It returns a generator, the next iteration in loop is executed in caller thread - in this case Main thread

src/neptune_fetcher/alpha/internal/attribute.py
michalsosn: to process each page concurrently, it could probably look the same way as https://github.com/neptune-ai/neptune-fetcher/pull/184/files#diff-8243a8fabb0b08a8d69e8035d604015fb30e7a3fd00cb91fe4016be9b74a49d5R104-R140

pitercl (approved):

PR #187, opened 2025-02-01
chore: Skip experiments filtering in list_attributes if filter is none

PatrykGala:


src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -189,13 +189,13 @@ def go_fetch_single(_filter: filter.AttributeFilter) -> Generator[util.Page[Attr
...
     project_identifiers: Iterable[identifiers.ProjectIdentifier],
-    experiment_identifiers: Iterable[identifiers.ExperimentIdentifier],
+    experiment_identifiers: Optional[Iterable[identifiers.ExperimentIdentifier]],
     attribute_filter: filter.AttributeFilter,
     batch_size: int,
 ) -> Generator[util.Page[AttributeDefinition], None, None]:
     params: dict[str, Any] = {
         "projectIdentifiers": list(project_identifiers),
-        "experimentIdsFilter": list(str(e) for e in experiment_identifiers),
+        "experimentIdsFilter": [str(e) for e in experiment_identifiers] if experiment_identifiers else None,
```
michalsosn, line 198: if experiment_identifiers is not None

michalsosn: otherwise you'll 
```
experiment_identifiers = None -> experimentIdsFilter: None -> return all experiments as expected
experiment_identifiers = [] -> experimentIdsFilter: None -> return all experiments, which is wrong
```

michalsosn: best to use `is None` if you expect sth to be None or not to avoid running into bugs with python 'truthiness'

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -409,11 +411,16 @@ def list_attributes(
     batch_size: int = env.NEPTUNE_FETCHER_ATTRIBUTE_DEFINITIONS_BATCH_SIZE.get(),
 ) -> Generator[str, None, None]:
     with util.create_thread_pool_executor() as executor:
+        if attribute_filter is None:
+            attribute_filter = filter.AttributeFilter()
+
+        def list_single_page(experiments_page: Optional[util.Page[experiment.ExperimentSysAttrs]]) -> set[str]:
+            experiment_identifiers = (
+                [identifiers.ExperimentIdentifier(project_id, e.sys_id) for e in experiments_page.items]
+                if experiments_page
```
michalsosn, line 420: `experiments_page` is a Page, it's probably always 'true'y', you probably want to check `experiments_page.items`

returning None, which means 'return all experiments', when the page is empty is likely wrong

src/neptune_fetcher/alpha/internal/attribute.py
PatrykGala: done

src/neptune_fetcher/alpha/internal/attribute.py
PatrykGala: done

michalsosn (approved):

PR #190, opened 2025-02-03
chore: not map values in threads

PatrykGala:

### DIFF
--- a/src/neptune_fetcher/alpha/internal/metric.py
+++ b/src/neptune_fetcher/alpha/internal/metric.py
@@ -44,7 +44,10 @@ from neptune_retrieval_api.models import (
     TimeSeries,
     TimeSeriesLineage,
 )
-from neptune_retrieval_api.proto.neptune_pb.api.v1.model.series_values_pb2 import ProtoFloatSeriesValuesResponseDTO
+from neptune_retrieval_api.proto.neptune_pb.api.v1.model.series_values_pb2 import (
+    ProtoFloatSeriesValuesDTO,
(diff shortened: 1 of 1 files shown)


src/neptune_fetcher/alpha/internal/metric.py
```diff
@@ -275,19 +294,23 @@ def _fetch_multiple_series_values(
 
         new_attribute_steps = {}
 
-        for path, series_values in values.items():
-            sorted_list = series_values if not tail_limit else reversed(series_values)
-            partial_results[path].extend(sorted_list)
+        for path, series_values in values:
+            sorted_values = series_values if not tail_limit else reversed(series_values)
```
pitercl, line 298: I thought about it some more and I think reversing might not be needed here as we sort the final DF by step anyway.

src/neptune_fetcher/alpha/internal/metric.py
```diff
@@ -248,7 +265,9 @@ def _fetch_multiple_series_values(
     tail_limit: Optional[int] = None,
 ) -> Iterable[_FloatPointValue]:
     results = []
-    partial_results: Dict[_AttributePathInExperiment, List[_FloatPointValue]] = {exp_path: [] for exp_path in exp_paths}
+    partial_results: Dict[_AttributePathInExperiment, Tuple[Iterable[_FloatPointValue], Length]] = {
```
pitercl, line 268: Not sure about `length` here, in increases the complexity of the code.
Maybe a small class that hides this? 
Not sure about perf impact, but I'd expect minimal.

PR #195, opened 2025-02-04
feat(alpha): rewrite list_attributes to use utils

michalsosn:


src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -406,46 +405,49 @@ def list_attributes(
...
+                items=util.split_experiments(
+                    [identifiers.ExperimentIdentifier(project_id, e.sys_id) for e in experiments_page.items]
+                ),
+                executor=executor,
+                downstream=lambda experiment_identifier_split: util.generate_concurrently(
+                    items=fetch_attribute_definitions(
+                        client,
+                        [project_id],
+                        experiment_identifier_split,
+                        cast(filter.AttributeFilter, attribute_filter),
```
pitercl, line 429: Remove the cast here and below

src/neptune_fetcher/alpha/internal/attribute.py
```diff
@@ -406,46 +405,49 @@ def list_attributes(
     project_id: identifiers.ProjectIdentifier,
     executor: Executor,
     fetch_attribute_definitions_executor: Executor,
-    experiment_filter: Optional[filter.Filter] = None,
-    attribute_filter: Optional[filter.BaseAttributeFilter] = None,
+    experiment_filter: Optional[filter.Filter],
```
pitercl, line 408: Could you pls move executors as the last 2 params? Seems weird having them in the middle of something semantically meaningful.

pitercl (changes requested):

PR #198, opened 2025-02-04
fix(alpha): check for conflict in attribute types when type_suffix_in_column_names=False

michalsosn:


src/neptune_fetcher/alpha/internal/output.py
```diff
@@ -68,13 +66,44 @@ def get_aggregation_subset(
...
+            (col[0].rsplit(":", 1)[0], col[1]) if isinstance(col, tuple) else col.rsplit(":", 1)[0]
+            for col in df.columns
+        ]
+
+        # Check for duplicate names
+        duplicated_names = df.columns[df.columns.duplicated(keep=False)]  # type: ignore
+        duplicated_names_set = set(duplicated_names)
+        if duplicated_names.any():
+            conflicting_types: dict[str, set[str]] = {}
+            for original_col, new_col in zip(original_columns, df.columns):
```
pitercl, line 85: This is fine, but I'd like to throw an exception defined in #196 and this info will not fit there.

pitercl (approved):

PR #207, opened 2025-02-06
feat(alpha): public runs api

michalsosn:


src/neptune_fetcher/alpha/__init__.py
```diff
@@ -24,14 +24,12 @@
     "list_attributes",
     "fetch_experiments_table",
     "fetch_metrics",
-    "filters",
```
PatrykGala, line 27: Why?

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -155,6 +161,19 @@ def fetch_metrics(
         return _transform_without_timestamp(df, type_suffix_in_column_names)
 
 
+def fetch_run_metrics(
+    experiments: Union[str, Filter],
```
PatrykGala, line 165: runs

PatrykGala, line 165:
```suggestion
    runs: Union[str, Filter],
```

src/neptune_fetcher/alpha/internal/composition/list_attributes.py
```diff
@@ -107,6 +110,14 @@ def list_attributes(
         return sorted(set(result))
 
 
+def list_run_attributes(
+    experiments: Optional[Union[str, Filter]] = None,
```
PatrykGala, line 114:
```suggestion
    runs: Optional[Union[str, Filter]] = None,
```

PatrykGala (changes requested):
experiments -> runs

src/neptune_fetcher/alpha/__init__.py
michalsosn: shouldn't filters be available simply because it's a module under alpha?
import neptune_fetcher.alpha.filters should work just because alpha/filters.py exists
or is some special effect intended here?

PR #208, opened 2025-02-06
chore: Add data for tests

PatrykGala:

### DIFF
--- a/tests/e2e/alpha/internal/generator.py
+++ b/tests/e2e/alpha/internal/generator.py
@@ -1,0 +1,228 @@
+from collections import defaultdict
+import concurrent.futures
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from optparse import Option
+from typing import Optional, Union
+import uuid
+import itertools
+from neptune_scale import Run
+from neptune_api import AuthenticatedClient
+
+from neptune_fetcher.alpha import Context, get_context
+from neptune_fetcher.alpha.internal.api_client import get_client
+import pytest
+
+METRICS_COUNT = 10
+POINTS_PER_METRIC = 1_000
+FORK_POINTS = [400, 800]
+RUN_COUNT = 5
+
+AttributeName = str
+Step = float
+Value = float
+
+
+@dataclass
+class GeneratedRun:
+    custom_run_id: str
+    experiment_name: str
+    fork_run_id: Union[str, None]
+    fork_level: Optional[int]
+    fork_point: Optional[int]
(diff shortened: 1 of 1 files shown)


tests/e2e/alpha/internal/generator.py
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    response = client.get_httpx_client().request(**args)
+    response.raise_for_status()
+
+
+@pytest.fixture(scope="session")
+def test_log_runs(client: AuthenticatedClient):
+    workspace = "neptune-e2e"  # TODO
+    project_name = f"pye2e-alpha-runs-{datetime.now().strftime('%Y_%m_%d_%H_%M_%S')}"
```
pitercl, line 193: Please add some random string at the end (5 chars max).

I'm also not super-convinced we should include `alpha` in the name. Doesn't matter much, can be easily changed later, but maybe we can think of this as a global setup and even if we write some "old" API tests (which will happen), we could use this?

Ideally I'd like all new tests to use this and we'll gradually move to use only this.

tests/e2e/alpha/internal/generator.py
```diff
@@ -0,0 +1,228 @@
...
+            "foo0": {800: 0.9, 900: 1.0, 1000: 1.1},
+            "foo1": {800: 1.9, 900: 2.0, 1000: 2.1},
+            "unique3/0": {800: 4.1, 900: 4.2, 1000: 4.3},
+        },
+    ),
+]
+
+
+def create_project(client: AuthenticatedClient, workspace: str, name: str):
+    body = {"organizationIdentifier": workspace, "name": name, "visibility": "priv"}
```
pitercl, line 179: Let's go with `workspace`. Less barriers.

pitercl (changes requested):

tests/e2e/alpha/internal/generator.py
```diff
@@ -0,0 +1,228 @@
...
+        fork_run_id=None,
+        configs={
+            "int-value": 1,
+            "float-value": 1.0,
+            "str-value": "hello_1",
+            "bool-value": False,
+            "datetime-value": datetime(2025, 1, 1, 1, 0, 0, 0, timezone.utc),
+        },
+        metrics={
+            "foo0": {0: 0.1, 100: 0.2, 200: 0.3, 300: 0.4, 400: 0.5},
```
pitercl, line 62: Btw, why such big steps? Any reason no to have 1, 2, 3, 4?

PR #209, opened 2025-02-07
feat(alpha): implement list_runs and list_attributes(runs)

michalsosn:


src/neptune_fetcher/alpha/internal/composition/attribute_components.py
```diff
@@ -0,0 +1,133 @@
...
+    client: AuthenticatedClient,
+    project_identifier: identifiers.ProjectIdentifier,
+    _filter: Optional[filters.Filter],
+    attribute_filter: filters.BaseAttributeFilter,
+    executor: Executor,
+    fetch_attribute_definitions_executor: Executor,
+    container_type: search.ContainerType,
+    downstream: Callable[[util.Page[att_defs.AttributeDefinition]], concurrency.OUT],
+) -> concurrency.OUT:
+    if container_type == search.ContainerType.RUN and _filter is None:
```
PatrykGala, line 103: Why only for runs?

src/neptune_fetcher/alpha/internal/composition/list_containers.py
```diff
@@ -48,33 +48,48 @@ def list_experiments(
...
+
+    return _list_containers(experiments, context, search.ContainerType.EXPERIMENT)
+
+
+def list_runs(
+    runs: Optional[Union[str, Filter]] = None,
+    context: Optional[_context.Context] = None,
+) -> list[str]:
+    if isinstance(runs, str):
+        runs = Filter.matches_all(Attribute("sys/custom_run_id", type="string"), regex=runs)  # TODO: custom_run_id?
```
PatrykGala, line 62:
```suggestion
        runs = Filter.matches_all(Attribute("sys/custom_run_id", type="string"), regex=runs) 
```

src/neptune_fetcher/alpha/internal/retrieval/search.py
```diff
@@ -49,47 +53,136 @@ def _map_direction(direction: Literal["asc", "desc"]) -> str:
...
+        container_type: ContainerType = ContainerType.EXPERIMENT,
+    ) -> Generator[util.Page[T], None, None]:
+        ...
+
+
+def _create_fetch_sys_attrs(
+    attribute_names: list[str],
+    make_record: Callable[[dict[str, Any]], T],
+    default_container_type: ContainerType,
+) -> FetchSysAttrs[T]:
```
PatrykGala, line 121: Is it correct return type?

src/neptune_fetcher/alpha/internal/retrieval/search.py
```diff
@@ -49,47 +53,136 @@ def _map_direction(direction: Literal["asc", "desc"]) -> str:
...
+            fetch_page=ft.partial(_fetch_sys_attrs_page, project_identifier=project_identifier),
+            process_page=ft.partial(_process_sys_attrs_page, make_record=make_record),
+            make_new_page_params=ft.partial(_make_new_sys_attrs_page_params, batch_size=batch_size, limit=limit),
+            params=params,
+        )
+
+    return fetch_sys_attrs
+
+
+fetch_experiment_sys_attrs = _create_fetch_sys_attrs(
```
PatrykGala, line 159: Can we use `ft.partial`?

src/neptune_fetcher/alpha/internal/composition/attribute_components.py
michalsosn: if I understand correctly, if _filter is None:

in case of runs, we can just invoke fetch_attribute_definitions with no experimentIdsFilter directly and it will return all unfiltered runs

in case of experiments, we still need to get ids of experiments (filter by experiment_leader=True), then fetch the attribute_definitions only for those ids

src/neptune_fetcher/alpha/internal/retrieval/search.py
michalsosn: it refers to this callable
class FetchSysAttrs(Protocol[T]):

I needed to define a protocol or mypy wouldn't know anything about named parameters that you can pass to a concrete fetch_sys_attrs

src/neptune_fetcher/alpha/internal/retrieval/search.py
michalsosn: it could work if we disable positional arguments completely

PatrykGala (approved):

PR #210, opened 2025-02-07
chore: add e2e for list_attributes and list_runs

PatrykGala:

### DIFF
--- a/tests/e2e/alpha/conftest.py
+++ b/tests/e2e/alpha/conftest.py
@@ -1,4 +1,6 @@
+import hashlib
 import os
+import platform
 import random
 import uuid
 from concurrent.futures import Executor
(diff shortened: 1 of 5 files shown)


tests/e2e/alpha/conftest.py
```diff
@@ -57,3 +65,61 @@ def random_series(length=10, start_step=0):
...
+    data.log_runs(project_id, ALL_STATIC_RUNS)
+    return project_id
+
+
+@fixture(scope="session")
+def new_project_context(new_project_id: str):
+    return get_context().with_project(new_project_id)
+
+
+def project_exists(client: AuthenticatedClient, workspace: str, project_id: str) -> bool:
```
michalsosn, line 101: I guess you could extract sending of the POST request as a separate function
def create_project(..) 
instead of defining it inside the @fixture def new_project_id()

tests/e2e/alpha/conftest.py
```diff
@@ -57,3 +65,61 @@ def random_series(length=10, start_step=0):
     values = [round((j + x) ** 3.0, 0) for x in range(len(steps))]
 
     return steps, values
+
+
+@fixture(scope="session")
+def workspace():
+    return NEPTUNE_E2E_WORKSPACE
```
michalsosn, line 72: hmm, why is NEPTUNE_E2E_WORKSPACE the only env/constant that you pass around as a fixture? 
e.g. you just use the value of NEPTUNE_E2E_REUSE_PROJECT directly
it doesn't really set up anything in the environment, so it's probably not really a fixture

tests/e2e/alpha/runs/test_runs_list.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.mark.parametrize(
+    "linear_history_filter",
+    [
+        "linear.*",
+        Filter.name_in(*[run.experiment_name for run in LINEAR_HISTORY_TREE]),
+        Filter.eq("linear-history", True),
+        Filter.eq(Attribute(name="linear-history", type="bool"), True),
+        Filter.eq(Attribute(name="linear-history", type="bool"), True),
```
michalsosn, line 32: isn't this case
`Filter.eq(Attribute(name="linear-history", type="bool"), True)`
duplicated? can't see a difference

michalsosn (dismissed):

michalsosn (approved):

PR #212, opened 2025-02-07
feat(alpha): implement fetch_metrics (runs)

michalsosn:


src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -289,37 +307,45 @@ def process_definitions(
...
                 )
-                _futures.append(executor.submit(process_definitions, sys_attrs_split, definitions_generator))
+                _futures.append(executor.submit(process_definitions, sys_id_labels_split, definitions_generator))
 
         return _futures, []
 
+    def fetch_sys_ids_labels() -> Generator[Iterable[SysIdLabel], None, None]:
+        if container_type == ContainerType.RUN:
+            run_pages = fetch_run_sys_attrs(client, project, _filter)
+            return ([SysIdLabel(run.sys_id, run.sys_custom_run_id) for run in page.items] for page in run_pages)
```
kgodlewski, line 340: Wouldn't it be better to yield `Iterable[Dict[SysId, str]]`? The order returned by the backend will be maintained, and we wouldn't need `SysIdLabel` at all, since we're using it to produce the dict in `process_sys_ids()`. So we'd save on intermediate dataclass construction.

```
return ({run.sys_id: run.sys_custom_run_id} for run in page.items for page in run_pages)
```

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -289,37 +307,45 @@ def process_definitions(
...
-            sys_ids = [exp.sys_id for exp in _experiments.items]
-            sys_id_to_sys_attrs = {exp.sys_id: exp for exp in _experiments.items}
+        if _sys_ids_labels:
+            sys_ids = [item.sys_id for item in _sys_ids_labels]
+            sys_id_to_labels = {item.sys_id: item for item in _sys_ids_labels}
 
             for sys_ids_split in split.split_sys_ids(sys_ids):
                 run_identifiers_split = [identifiers.RunIdentifier(project, sys_id) for sys_id in sys_ids_split]
-                sys_attrs_split = [sys_id_to_sys_attrs[sys_id] for sys_id in sys_ids_split]
+                sys_id_labels_split = [sys_id_to_labels[sys_id] for sys_id in sys_ids_split]
```
kgodlewski, line 325: IMO `sys_id_labels_batches` would be a bit more informative

michalsosn: idk, batching means we start with individual items and add them into a collection
in split_sys_ids we start with a batch and split it into smaller batches that fit into the API's limits

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -109,40 +111,88 @@ def fetch_experiment_metrics(
...
     If `include_time` is set, each metric column has an additional sub-column with requested timestamp values.
     """
+    if isinstance(experiments, str):
+        experiments = Filter.matches_all(Attribute("sys/name", type="string"), regex=experiments)
+
+    if isinstance(attributes, str):
+        attributes = AttributeFilter(name_matches_all=attributes, type_in=["float_series"])
+
+    return _fetch_metrics(
+        _filter=experiments,
```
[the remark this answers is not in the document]
michalsosn: Well, this PR went from changing 2 files to changing 15 files but sure

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
michalsosn: removing SysIdLabel sounds good

kgodlewski (approved):

PR #214, opened 2025-02-10
Add docstrings for Filter classes

normandy7:
Add docstrings for `Filter`, `Attribute`, and `AttributeFilter`.

- [x] Did you **ask the docs owner** to review all the user-facing changes?

### DIFF
--- a/src/neptune_fetcher/alpha/filters.py
+++ b/src/neptune_fetcher/alpha/filters.py
@@ -138,25 +138,15 @@ class Attribute:
 
     Examples:
 
-    Import the needed classes:
+    Select a metric and pick variance as the aggregation:
 
     ```
     import neptune_fetcher.alpha as npt
     from npt.filters import Attribute, Filter
-    ```
 
-    Select attribute by exact name:
 
-    ```
-    optimizer = Attribute(name="config/optimizer")
-    ```
-
-    Select a metric and pick variance as the aggregation:
-
-    ```
     val_loss_variance = Attribute(
         name="val/loss",
-        type=["float_series"],  # ensures that the attribute is a numerical series
         aggregation="variance",
     )
     ```


src/neptune_fetcher/alpha/filters.py
```diff
@@ -69,6 +122,51 @@ def __or__(self, other: "BaseAttributeFilter") -> "BaseAttributeFilter":
 
 @dataclass
 class Attribute:
+    """Helper for specifying an attribute and picking a metric aggregation function.
+
+    Use to build filters for fetching experiments or runs.
```
pitercl, line 127: It's actually used in one other place: in `fetch_{runs|experiments}_table` you can specify a `sort_by` attribute.

src/neptune_fetcher/alpha/filters.py
```diff
@@ -69,6 +122,51 @@ def __or__(self, other: "BaseAttributeFilter") -> "BaseAttributeFilter":
...
+
+    Args:
+        name (str): An attribute name to match exactly.
+        aggregation (Literal["last", "min", "max", "average", "variance"], optional):
+            Aggregation function to apply when specifying a metric of type FloatSeries.
+            Defaults to `"last"`, i.e. the last logged value.
+        type (Literal["float", "int", "string", "bool", "datetime", "float_series", "string_set"], optional):
+            Allowed attribute type. For a reference, see: https://docs-beta.neptune.ai/attribute_types
```
pitercl, line 135: It's not as much `Allowed attribute type` as just `Attribute type`. This arg enables users to explicitly state the type in case the type cannot be inferred automatically, or they want to write future-proof script (i.e. even if now the type can be inferred, in the future it might become ambiguous if some experiment/run adds an attribute of the same name but different type).

src/neptune_fetcher/alpha/filters.py
```diff
@@ -69,6 +122,51 @@ def __or__(self, other: "BaseAttributeFilter") -> "BaseAttributeFilter":
...
+
+    ```
+    import neptune_fetcher.alpha as npt
+    from npt.filters import Attribute, Filter
+    ```
+
+    Select attribute by exact name:
+
+    ```
+    optimizer = Attribute(name="config/optimizer")
```
pitercl, line 149: While technically correct, I'm not sure if it makes sense in this doc string. It doesn't differ from just `"config/optimizer"`.

src/neptune_fetcher/alpha/filters.py
```diff
@@ -69,6 +122,51 @@ def __or__(self, other: "BaseAttributeFilter") -> "BaseAttributeFilter":
...
+    ```
+
+    Select a metric and pick variance as the aggregation:
+
+    ```
+    val_loss_variance = Attribute(
+        name="val/loss",
+        type=["float_series"],  # ensures that the attribute is a numerical series
```
pitercl, line 157: `type` is a string, not a list.

And, btw, if `aggregation` is specified, type will be automatically inferred to be `float_series`. Not sure if it's an important detail, but maybe worth mentioning :shrug:

pitercl (changes requested):

src/neptune_fetcher/alpha/filters.py
normandy7: How about:

```suggestion
    When fetching experiments or runs, use this class to filter and sort the returned entries.
```

pitercl: I'd say it's ok.
I just realized it's actually used when fetching metrics, too. But since we're fetching metrics for runs/exps, maybe it's ok?

src/neptune_fetcher/alpha/filters.py
normandy7: I see. For clarity, we should spell it out:

```suggestion
            Attribute type. Specify it to resolve ambiguity, in case some of the project's runs contain attributes
            that have the same name but are of a different type.
            For a reference, see: https://docs-beta.neptune.ai/attribute_types
```

src/neptune_fetcher/alpha/filters.py
normandy7: Ah ok. Then I think we can just omit `type` here. Do we print something helpful if people try to get an aggregation of something other than a float series?

src/neptune_fetcher/alpha/filters.py
normandy7: Yeah, the semantics get kinda blurred. I don't think the phrasing here will make or break anything.

src/neptune_fetcher/alpha/filters.py
pitercl: No. We just assume it's a float series and query the API.
```python
list_experiments(experiments=Filter.gt(Attribute("not a series", aggregation="max"), 0))
``` 
This will return an empty list; exactly as if the attribute of such name didn't exist at all in the project.

normandy7: Right. If specifying `type="float_series"` won't make a difference w/rt this behavior, I think we don't need it in this example.

PR #215, opened 2025-02-10
docs(alpha): add docstrings to functions reading runs

michalsosn:


src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -140,6 +140,28 @@ def fetch_run_metrics(
...
+                If `AttributeFilter.aggregations` is set, an exception will be raised as
+                they're not supported in this function.
+    `include_time` - whether to include absolute timestamp
+    `step_range` - a tuple specifying the range of steps to include; can represent an open interval
+    `lineage_to_the_root` - if True (default), includes all points from the complete run history.
+        If False, only includes points from the most recent run in the lineage.
+    `tail_limit` - from the tail end of each series, how many points to include at most.
+    `type_suffix_in_column_names` - False by default. If True, columns of the returned DataFrame
+        will be suffixed with ":<type>", e.g. "attribute1:float_series", "attribute1:string", etc.
+        If set to False, the method throws an exception if there are multiple types under one path.
```
normandy7, line 161:
```suggestion
    `type_suffix_in_column_names` - False by default. If set to True, columns of the returned DataFrame
        are suffixed with ":<type>", e.g. "attribute1:float_series", "attribute1:string".
        If False, an exception is raised if there are multiple types under one attribute path.
```

src/neptune_fetcher/alpha/internal/composition/fetch_table.py
```diff
@@ -111,6 +111,26 @@ def fetch_runs_table(
...
+        - an AttributeFilter object
+    `sort_by` - an attribute name or an Attribute object specifying type and, optionally, aggregation
+    `sort_direction` - 'asc' or 'desc'
+    `limit` - maximum number of runs to return; by default all runs are returned.
+    `type_suffix_in_column_names` - False by default. If True, columns of the returned DataFrame
+        will be suffixed with ":<type>", e.g. "attribute1:float_series", "attribute1:string", etc.
+        If set to False, the method throws an exception if there are multiple types under one path.
+    `context` - a Context object to be used; primarily useful for switching projects
+
+    Returns a DataFrame similar to the Experiments Table in the UI, with an important difference:
```
normandy7, line 129:
```suggestion
    Returns a DataFrame similar to the runs table in the web app, with an important difference:
```

pitercl (approved):

PR #222, opened 2025-02-11
fix(alpha): add additional project validation and ValidContext

michalsosn:
- refactor code so that mypy is aware when the context has been validated and can be used without # type: ignore
- add _PROJECT_PATH_PATTERN to reject some wrong project names before a request is sent


src/neptune_fetcher/alpha/internal/context.py
```diff
@@ -110,15 +117,24 @@ def set_context(context: Optional[Context] = None) -> Context:
         return _context
 
 
-def validate_context(context: Optional[Context] = None) -> Context:
+def get_valid_context(context: Optional[Context] = None) -> ValidContext:
+    return _validate_context(context or get_context())
+
+
+_PROJECT_PATH_PATTERN = re.compile(r"^([A-Za-z0-9-._]+)/([^/\\#?%]+)$")
```
PatrykGala, line 124: project name can contains utf8 characters: 
https://github.com/neptune-ai/neptune-fetcher/blob/main/.github/scripts/run-e2e-tests.sh#L21

michalsosn: and `[^/\\#?%]` matches all characters except /\\#?%, so they're covered, no?

michalsosn: added some tests cases with unicode chars

tests/unit/alpha/internal/test_api_client_cache.py
```diff
@@ -79,45 +79,45 @@ def mock_networking():
 def test_same_token(context):
     """Should return a single instance of the client because the token is the same"""
 
-    client = get_client(context)
-    assert get_client(context) is client
+    client = get_client(context.api_token)
```
pitercl, line 82: If you're set on changing the API of the client (I'm not strongly against, but I don't see the need), please change this test, too - context is unnecessary here.

src/neptune_fetcher/alpha/internal/context.py
pitercl: This regex provides only partial validation: backend accepts 3-byte UTF-8 chars, but not 4-byte chars in the project name. This regex matches the 4-byte chars, too.

And so, I don't like it - we can add it if you fix the regex.

pitercl (changes requested):

src/neptune_fetcher/alpha/internal/context.py
michalsosn: changed the regex to be equal to https://gitlab.com/neptune.ml/neptune/-/blob/main/apps/ingestion-gateway/src/service/ingest/validate.rs?ref_type=heads#L11

michalsosn:
my comment got deleted from some reason...
but I extended the regex to accept more characters - found one in the backend code. The link was in the deleted comment, but it wasn't difficult to find

PR #229, opened 2025-02-14
feat: Implement include_preview

PatrykGala:

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -7,4 +7,4 @@ pytest-mock
 pytest-retry
 pytest-timeout
 icecream
-neptune-scale==0.9.1b2
+git+https://github.com/neptune-ai/neptune-client-scale.git@main
--- a/src/neptune_fetcher/fields.py
+++ b/src/neptune_fetcher/fields.py
@@ -104,7 +104,7 @@ class FloatSeries:
         include_inherited: bool = True,
         progress_bar: bool = True,
         step_range: Tuple[Union[float, None], Union[float, None]] = (None, None),
-        include_preview: bool = False,
+        include_point_previews: bool = False,
     ) -> "DataFrame":
         import pandas as pd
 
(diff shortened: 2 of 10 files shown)


src/neptune_fetcher/fields.py
```diff
@@ -102,18 +104,27 @@ def fetch_values(
...
+                include_inherited=include_inherited,
+                step_range=step_range,
+                include_preview=include_preview,
             )
 
         else:
             data = self.prefetched_data
 
-        rows = dict((n, make_row(entry=entry, include_timestamp=include_timestamp)) for (n, entry) in enumerate(data))
+        filtered_data = (entry for entry in data if include_preview or not entry.preview)
```
michalsosn, line 123: hmm, when is it possible that include_preview = False and entry.preview = True?

michalsosn: ah, I guess when you use prefetched_data, which includes previews, but fetch_values is called with include_preview=False

src/neptune_fetcher/fields.py
```diff
@@ -171,4 +184,8 @@ def make_row(entry: FloatPointValue, include_timestamp: bool = True) -> Dict[str
     if include_timestamp:
         row["timestamp"] = entry.timestamp
 
+    if include_preview:
+        row["preview"] = entry.preview
+        row["completion_ratio"] = entry.completion_ratio
```
michalsosn, line 189: is preview == True <=> completion_ratio == 1.0 ?

MaciekPytel: preview == true -> completion_ration == 1.0, but the implication only goes one way. A preview with completion_ratio of 1.0 is _not_ the same as committed point.
Idea is that the boolean value determines the system behavior. Completion is just additional metadata that user can specify for preview points (and they can interpret it any way they want). Neptune doesn't infer anything from it or modify behavior in any way. All of the below are valid:
 * Writing preview with completion of 1.0.
 * Overriding a preview with completion of 0.5 with a preview with the same step with completion of 0.2.
 * Overriding a preview with completion of 0.2 with a preview with the same step with completion of 0.5.

michalsosn (approved):

dev_requirements.txt
```diff
@@ -7,4 +7,4 @@ pytest-mock
 pytest-retry
 pytest-timeout
 icecream
-neptune-scale==0.9.1b2
+git+https://github.com/neptune-ai/neptune-client-scale.git@main
```
michalsosn, line 10: this is not temporary but an improvement I guess

michalsosn (approved):

src/neptune_fetcher/fields.py
```diff
@@ -114,15 +114,23 @@ def fetch_values(
...
             data = self.prefetched_data
 
-        filtered_data = (entry for entry in data if include_preview or not entry.preview)
+        # The motivation for this is to filter-out preview points downloaded during prefetching
+        # in case the user does not request them.
+        filtered_data = (entry for entry in data if include_point_previews or not entry.preview)
+
         rows = dict(
-            (n, make_row(entry=entry, include_timestamp=include_timestamp, include_preview=include_preview))
+            (
```
kgodlewski, line 128: Using a dict comprehension here, instead of creating intermediate tuples will be more efficient (~9% faster based on a quick test)

PR #230, opened 2025-02-17
chore(alpha): set dataclasses to use slots

michalsosn:


michalsosn:
maybe when we drop support for python 3.9 because slots were added later

PR #231, opened 2025-02-17
Raise `NeptuneProjectInaccessible` on "access denied" server response

kgodlewski:


src/neptune_fetcher/alpha/exceptions.py
```diff
@@ -54,6 +54,17 @@ def __init__(self) -> None:
         )
 
 
+class NeptuneProjectInaccessible(NeptuneError):
+    def __init__(self) -> None:
+        super().__init__(
+            """
+{h1}NeptuneProjectInaccessible: The project is not accessible or does not exist.{end}
```
michalsosn, line 61: could we add the current project as an argument?
`The project '{project}' is not accessible`
It could sometimes help the user figure out an error
project is not really a secret, so it should be ok to log it

michalsosn: hmm, I guess it won't be easy because backoff_retry is too generic to easily get the project that was passed
nevermind then

michalsosn (approved):

src/neptune_fetcher/alpha/exceptions.py
```diff
@@ -54,6 +54,17 @@ def __init__(self) -> None:
...
+
+Ensure the project name is correct and that you have the necessary access permissions.
```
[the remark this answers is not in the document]
kgodlewski: This error actually captures a case where either the project or workspace do not exist, or the user has no viewer access (or better)

normandy7 (approved):

pitercl (commented):
.

pitercl (changes requested):
.

src/neptune_fetcher/alpha/internal/retrieval/util.py
```diff
@@ -122,3 +121,30 @@ def backoff_retry(
...
+    The content is assumed to be valid JSON and contain the "errorType" key.
+    If it is not the case, or the response does not match any known exceptions,
+    `NeptuneUnexpectedResponseError` is raised.
+    """
+    try:
+        if content is not None:
+            json_content = json.loads(content.decode("utf-8"))
+
+            error_type = json_content.get("errorType")
+            if error_type == "ACCESS_DENIED":
```
pitercl, line 139: Why not use status code? it's far more likely someone in the backend changes this string than status code.

src/neptune_fetcher/alpha/internal/retrieval/util.py
```diff
@@ -122,3 +121,30 @@ def backoff_retry(
         raise error from last_exc
     else:
         raise error
+
+
+def raise_for_response(status_code: int, content: bytes) -> Exception:
```
pitercl, line 126: I'd prefer this to be an internal function of this module (i.e. `_raise_for_response`). I wouldn't want this module to become a bag for everything that doesn't clearly fit elsewhere. This function is an impl detail of retry backoff, at least for now - not an attempt at some global error handling.

pitercl (changes requested):

pitercl (approved):

PR #234, opened 2025-02-18
Don't crash on user metric names that conflict with our DF columns.

kgodlewski:
This is done by operating on an intermediate DataFrame that has int columns representing user metrics, instead of the actual names.

Once the DF index is recreated, we restore the original column names and proceed as before.


src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -241,20 +241,20 @@ def _fetch_metrics(
 
         index_column_name = "experiment" if container_type == ContainerType.EXPERIMENT else "run"
```
pitercl, line 242: Please inline this

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -391,14 +391,39 @@ def _start() -> Iterable[FloatPointValue]:
...
 
 
-def _create_flat_dataframe(
-    values: Iterable[FloatPointValue],
+def _create_dataframe(
+    data_points: Iterable[FloatPointValue],
+    *,
+    type_suffix_in_column_names: bool,
     include_point_previews: bool,
     index_column_name: str = "experiment",
```
pitercl, line 399: Pls remove the default for `index_column_name`.

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -391,14 +391,39 @@ def _start() -> Iterable[FloatPointValue]:
...
+    timestamp_column_name: Optional[str] = None,
 ) -> pd.DataFrame:
     """
-    Creates a memory-efficient DataFrame directly from _FloatPointValue tuples
+    Creates a memory-efficient DataFrame directly from FloatPointValue tuples
     by converting strings to categorical codes before DataFrame creation.
+
```
pitercl, line 405: This comment is no longer true, but in general I think it makes sense to have a comment on what this function does - it's not obvious.

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -444,73 +469,66 @@ def generate_categorized_rows(float_point_values: Iterable[FloatPointValue]) ->
...
+        if include_point_previews:
+            values.append("is_preview")
+            values.append("preview_completion")
 
+    df = df.pivot(index=[index_column_name, "step"], columns="path", values=values)
     df = df.reset_index()
     df[index_column_name] = df[index_column_name].astype(str)
     df = df.sort_values(by=[index_column_name, "step"], ignore_index=True)
-    df.columns.names = (None, None)
     df = df.set_index([index_column_name, "step"])
```
michalsosn, line 516:
```
    df = df.reset_index()
    df[index_column_name] = df[index_column_name].astype(str)
    df = df.sort_values(by=[index_column_name, "step"], ignore_index=True)
    df = df.set_index([index_column_name, "step"])
```
->
```
    df.reset_index(inplace=True)
    df[index_column_name] = df[index_column_name].astype(str)
    df.sort_values(by=[index_column_name, "step"], ignore_index=True, inplace=True)
    df.set_index([index_column_name, "step"], inplace=True)
```
would probably avoid copying the dataframe 3 times

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -444,73 +469,66 @@ def generate_categorized_rows(float_point_values: Iterable[FloatPointValue]) ->
...
-        columns="path",
-        values=values,
-    )
-
-    df = df.swaplevel(axis=1)
-    if type_suffix_in_column_names:
-        df = df.rename(columns=lambda x: x + ":float_series", level=0, copy=False)
+    values: Union[str, list[str]] = "value"
+
+    # Create multi-index if necessary
```
michalsosn, line 502: # Create column multi-index if necessary

There are two indices - for rows and columns
and as you can see in `df.pivot(index=[index_column_name, "step"], columns="path", values=values)`
usually when you don't specify which one you mean it's assumed that you mean the row index, which is not the case here

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -444,73 +469,66 @@ def generate_categorized_rows(float_point_values: Iterable[FloatPointValue]) ->
...
+    if include_point_previews or timestamp_column_name:
+        values = ["value"]
+        if timestamp_column_name:
+            df[timestamp_column_name] = pd.to_datetime(df[timestamp_column_name], unit="ms", origin="unix", utc=True)
+            values.append(timestamp_column_name)
+        if include_point_previews:
+            values.append("is_preview")
+            values.append("preview_completion")
```
michalsosn, line 510:
```
    values: Union[str, list[str]] = "value"

    # Create multi-index if necessary
    if include_point_previews or timestamp_column_name:
        values = ["value"]
        if timestamp_column_name:
            df[timestamp_column_name] = pd.to_datetime(df[timestamp_column_name], unit="ms", origin="unix", utc=True)
            values.append(timestamp_column_name)
        if include_point_previews:
            values.append("is_preview")
            values.append("preview_completion")
```
->
```
    values = ["value"]
    # Create multi-index if necessary
    if timestamp_column_name:
        df[timestamp_column_name] = pd.to_datetime(df[timestamp_column_name], unit="ms", origin="unix", utc=True)
        values.append(timestamp_column_name)
    if include_point_previews:
        values.append("is_preview")
        values.append("preview_completion")
```
Can't you just start with a list?

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -444,73 +469,66 @@ def generate_categorized_rows(float_point_values: Iterable[FloatPointValue]) ->
     types = [
         (index_column_name, "uint32"),
         ("path", "uint32"),
-        ("timestamp", "uint64"),
+        (timestamp_column_name or "timestamp", "uint64"),
```
pitercl, line 472: I think we should adjust if `generate_categorized_rows` produces the timestamp column based on whether we actually need it.

michalsosn, line 472: if timestamp_column_name is None means that we the result dataframe won't have this column at all, then shouldn't we already exclude it at this point as we do with `if include_point_previews:` ? 

```
if timestamp_column_name is None:
    types.append((timestamp_column_name, "uint64"))
```

pitercl (changes requested):

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -241,20 +241,20 @@ def _fetch_metrics(
...
-    if include_time == "absolute":
-        df = _transform_with_absolute_timestamp(
-            df, type_suffix_in_column_names, include_point_previews, index_column_name
-        )
-    # elif include_time == "relative":
-    #     raise NotImplementedError("Relative timestamp is not implemented")
+    # MultiIndex DFs need to have column index order swapped: value/metric_name -> metric_name/value
+    if isinstance(df.columns, pd.MultiIndex):
+        df = df.swaplevel(axis=1)
+        df = df.sort_index(axis=1, level=0)
```
michalsosn, line 255:
```
df = df.sort_index(axis=1)
```
Wouldn't not specifying level=0 sort both levels instead of leaving the 2nd one unsorted?

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
kgodlewski: No, as this will create a column multiindex: `[foo/value, bar/value] instead of [foo, bar]`. This is the behaviour we currently have -- no multiindex in the common case.

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
kgodlewski: The column index, I'll clarify that

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
```diff
@@ -391,14 +391,39 @@ def _start() -> Iterable[FloatPointValue]:
     return _start()
 
 
-def _create_flat_dataframe(
```
michalsosn, line 394: btw. I'd move _create_dataframe to internal.output_format
https://github.com/neptune-ai/neptune-fetcher/blob/main/src/neptune_fetcher/alpha/internal/output_format.py
This function is tested directly in unit tests despite being private, so it's a sign that is could probably be a public entry point in a different module

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
kgodlewski: It would, I left it as it was implemented before -- apparently the intent was to only sort by metric name, leaving the 2nd-level index as is

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
kgodlewski: We could exclude it, but it'll make the row generation logic more complex, as we will need to dynamically adjust the generated row length to the mix of timestamp and preview columns.

src/neptune_fetcher/alpha/internal/composition/fetch_metrics.py
kgodlewski: We'll benchmark this separately and if there's an improvement, a separate PR will be created

michalsosn (approved):

tests/unit/alpha/test_fetch_metrics.py
```diff
@@ -1,56 +1,121 @@
-from datetime import datetime
+import itertools
```
pitercl, line 1: Please rename this file - it doesn't test fetch_metrics; it tests output format.

tests/e2e/alpha/runs/test_runs_fetch_metrics.py
```diff
@@ -213,20 +210,12 @@ def create_expected_data(expected_metrics, include_time: str, type_suffix_in_col
...
 
-    if include_time == "absolute":
-        return _transform_with_absolute_timestamp(
-            df, type_suffix_in_column_names, include_point_previews=False, index_column_name="run"
-        )
-    else:
-        return _transform_without_timestamp(
-            df, type_suffix_in_column_names, include_point_previews=False, index_column_name="run"
-        )
+    return create_dataframe(
```
pitercl, line 215: For the record, I don't like that we're using prod code here - we'll change it. But only a slight regression, so let's not do it in this PR.

tests/e2e/alpha/internal/test_fetch_metrics.py
```diff
@@ -162,32 +159,27 @@ def create_expected_data(
...
             limited = filtered[-tail_limit:] if tail_limit is not None else filtered
             rows.extend(limited)
 
-    df = pd.DataFrame(rows, columns=["experiment", "path", "timestamp", "step", "value"])
-    df["experiment"] = df["experiment"].astype(str)
-    df["path"] = df["path"].astype(str)
-    df["timestamp"] = df["timestamp"].astype(int)
-    df["step"] = df["step"].astype(float)
-    df["value"] = df["value"].astype(float)
+    df = create_dataframe(
```
pitercl, line 169: For the record, I don't like that we're using prod code here - we'll change it. But only a slight regression, so let's not do it in this PR.

pitercl (changes requested):

michalsosn (approved):

pitercl (approved):

PR #237, opened 2025-02-25
feat(alpha): string_series 

michalsosn:


michalsosn:
[searching experiments by nql] search_leaderboard_entries_proto
Searching by attribute of type STRING_SERIES never returns results.
Invalid aggregations: LAST_STEP and SIZE. LAST is valid, but does not return anything anyway.

[attribute definitions] query_attribute_definitions_within_project
works

[downloading attribute values] query_attributes_within_project_proto
Fails with:
    2025-04-11 03:32:33,558 ERROR [administrator] [http-nio-8088-exec-3] m.n.l.a.h.GenericExceptionHandler: Unhandled exception occurred.java.lang.IllegalArgumentException: Unsupported type: STRING_SERIES
    at ml.neptune.leaderboard.infrastructure.database.repositories.fields.ProjectRunFieldRepositoryImpl.deserializeFieldRows(ProjectRunFieldRepositoryImpl.kt:329)
    ...
    at ml.neptune.leaderboard.api.controllers.leaderboard.ProtoLeaderboardAttributesApiController$$EnhancerBySpringCGLIB$$2908c8a2.queryAttributesWithinProjectProto(<generated>)

[downloading series] get_series_values_proto
Returns no points, even though there are segments in the db
Returned SearchAfter has finished=False but token=""  (token is an empty string ??)

michalsosn:
get_series_values_proto started returning results after the fix in https://github.com/neptune-ai/neptune-client-scale/pull/202/files

src/neptune_fetcher/alpha/internal/composition/fetch_series.py
```diff
@@ -0,0 +1,181 @@
...
+                downstream=lambda sys_ids_split, definitions_page: concurrency.generate_concurrently(
+                    items=series.fetch_series_values(
+                        client=client,
+                        run_attribute_definitions=[
+                            series.RunAttributeDefinition(
+                                run_identifier=identifiers.RunIdentifier(project_identifier, sys_id),
+                                attribute_definition=definition,
+                            )
+                            for sys_id in sys_ids_split
+                            for definition in _filter_series(definitions_page.items)
```
pitercl, line 116: Shouldn't we modify the `attributes` filter to be sure that we're only downloading types we need? Or maybe rather copy it and modify the copy.

src/neptune_fetcher/alpha/internal/retrieval/series.py
```diff
@@ -0,0 +1,150 @@
...
+    run_attribute_definitions: Iterable[RunAttributeDefinition],
+    include_inherited: bool,
+    step_range: Tuple[Union[float, None], Union[float, None]] = (None, None),
+    tail_limit: Optional[int] = None,
+) -> Generator[util.Page[tuple[RunAttributeDefinition, list[StringSeriesValue]]], None, None]:
+    if not run_attribute_definitions:
+        yield from []
+        return
+
+    request_id_to_runs_definitions: dict[str, RunAttributeDefinition] = {
```
pitercl, line 57: `request_id_to_run_attribute_definition`

pitercl (changes requested):
Haven't seen the tests yet.

tests/e2e/alpha/internal/test_fetch_series.py
```diff
@@ -0,0 +1,213 @@
...
+                filter_=Filter.name_in(*expected_experiments),
+            )
+        ]
+    )
+    if expected_experiments == {exp.sys_name for exp in existing_experiments}:
+        return {}
+
+    runs = {}
+    for experiment in TEST_DATA.experiments:
+        run = Run(
```
pitercl, line 104: Why don't we use the globally defined data and only fetch it here?

tests/e2e/alpha/internal/retrieval/test_series.py
```diff
@@ -0,0 +1,189 @@
...
+            identifiers.ProjectIdentifier(project_identifier),
+            Filter.name_in(EXPERIMENT_NAME),
+        )
+    )
+    if existing.items:
+        return
+
+    run_id = str(uuid.uuid4())
+
+    run = Run(
```
pitercl, line 47: Let's use global data, pls

pitercl (changes requested):

tests/e2e/alpha/internal/test_fetch_series.py
michalsosn: ...outside of the fact that they all have slightly different data and I have to fix assertions sure

michalsosn: rewrote everything except from inference tests, because they are somewhat unique

michalsosn:
there are two tests with mark.skip, waiting for the change on the backend

src/neptune_fetcher/alpha/internal/composition/fetch_series.py
```diff
@@ -0,0 +1,186 @@
...
+def _validate_tail_limit(tail_limit: Optional[int]) -> None:
+    """Validate that tail_limit is either None or a positive integer."""
+    if tail_limit is not None:
+        if not isinstance(tail_limit, int):
+            raise ValueError("tail_limit must be None or an integer")
+        if tail_limit <= 0:
+            raise ValueError("tail_limit must be greater than 0")
+
+
+def _filter_series(
```
pitercl, line 183: Since we now have an assert, maybe we can drop this? I'd prefer to avoid defensive programming.

michalsosn: sure, there's just a chance that if AttributeFilter ever becomes anything more advanced than just a name_id / type_in expression, someone will have to remember about restoring it

pyproject.toml
```diff
@@ -12,7 +12,7 @@ pattern = "default-unprefixed"
 python = "^3.9"
 
 # Base neptune package
-neptune-api = {git = "https://github.com/neptune-ai/neptune-api", rev="kg/new-swagger-code"}
+neptune-api = {git = "https://github.com/neptune-ai/neptune-api", rev="ms/gen-15"}
```
pitercl, line 15: This needs to be changed before merging

src/neptune_fetcher/alpha/internal/composition/attributes.py
```diff
@@ -91,6 +92,9 @@ def fetch_attribute_definition_aggregations(
 
             if definition.type == "float_series" or definition.type == "string_series":
```
pitercl, line 93: Could we change it to `if definition.type in TYPE_AGGREGATIONS.keys()`? Seems more resilient given the `if` below.

pitercl (changes requested):

PR #256, opened 2025-03-31

michalsosn:


michalsosn:
this one... affects public docs, but I'll wait until [ms/fix-docstrings](https://github.com/neptune-ai/neptune-fetcher/tree/ms/fix-docstrings) is merged before I ask for a review

PR #257, opened 2025-03-31
Add `NeptuneInvalidCredentialsError`

kgodlewski:
The new error is raised when the backend explicitly rejects the API token as invalid (as opposed to any other token exchange error, eg. HTTP 500). We don't retry requests when this happens, but raise the error immediately instead.

This PR depends on changes in `neptune-api`: https://github.com/neptune-ai/neptune-api/pull/74

### DIFF
--- a/src/neptune_fetcher/alpha/exceptions.py
+++ b/src/neptune_fetcher/alpha/exceptions.py
@@ -80,6 +80,19 @@ or with the `NEPTUNE_API_TOKEN` environment variable.
         )
 
 
+class NeptuneInvalidCredentialsError(NeptuneUserError):
+    def __init__(self) -> None:
+        super().__init__(
+            """
+{h1}NeptuneInvalidCredentialsError: Your Neptune API token was rejected by the server.{end}
+
+Make sure to specify a valid API token with `set_api_token()` function,
+by providing the `context` argument with `project` and `api_token` set to the function call,
+or with the `NEPTUNE_API_TOKEN` environment variable.
+"""
+        )
+
+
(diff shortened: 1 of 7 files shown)


src/neptune_fetcher/alpha/exceptions.py
```diff
@@ -80,6 +80,19 @@ def __init__(self) -> None:
...
+class NeptuneInvalidCredentialsError(NeptuneUserError):
+    def __init__(self) -> None:
+        super().__init__(
+            """
+{h1}NeptuneInvalidCredentialsError: Your Neptune API token was rejected by the server.{end}
+
+Make sure to specify a valid API token with `set_api_token()` function,
+by providing the `context` argument with `project` and `api_token` set to the function call,
+or with the `NEPTUNE_API_TOKEN` environment variable.
+"""
```
normandy7, line 92:
```suggestion
Make sure to specify a valid token in one of the following ways:

- Call the `set_api_token()` function
- Create a Context with the API token and pass it to the `context` argument of the fetching method
- Set the `NEPTUNE_API_TOKEN` environment variable

For details, see https://docs-beta.neptune.ai/fetcher_setup
"""
```

kgodlewski: @normandy7 can you have a look at `NeptuneApiTokenNotProvided`? This is where I got the initial message from -- should I update there too?

normandy7: Yes. Does this work? → #261

PR #259, opened 2025-04-01
Add NEPTUNE_HTTP_REQUEST_TIMEOUT_SECONDS

PatrykGala:
Add a new env variable `NEPTUNE_HTTP_REQUEST_TIMEOUT_SECONDS` that controls `httpx` timeouts. Defaults to 60 sec.

### DIFF
--- a/src/neptune_fetcher/api/api_client.py
+++ b/src/neptune_fetcher/api/api_client.py
@@ -21,17 +21,13 @@ __all__ = (
     "get_config_and_token_urls",
 )
 
-import logging
 import re
-import time
 from dataclasses import dataclass
 from datetime import (
     datetime,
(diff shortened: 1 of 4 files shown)


src/neptune_fetcher/api/api_client.py
```diff
@@ -336,6 +337,14 @@ def backoff_retry(
             raise NeptuneException(
                 "Your API token was rejected by the Neptune backend because it is either unknown or expired."
             ) from e
+        except httpx.TimeoutException as e:
```
michalsosn, line 340: hmm, there's a duplicate backoff_retry in alpha.internal.retrieval.util

michalsosn (approved):

PR #265, opened 2025-04-02

michalsosn:


pitercl (approved):

pitercl (changes requested):
Wait, we need to update the docs as well

michalsosn:
should the docs be updated now? I'm merging everything to dev/move-alpha branch, so I could leave a main->dev/move-alpha PR open this time and wait for docs at this stage

PR #269, opened 2025-04-07
chore(alpha): move alpha to the top level

michalsosn:


tests/unit/legacy/test_api_client.py
```diff
@@ -19,8 +19,8 @@
 )
 from pytest import fixture
 
-from neptune_fetcher.api.api_client import ApiClient
-from neptune_fetcher.fields import FloatPointValue
+from neptune_fetcher.legacy.api.api_client import ApiClient
```
pitercl, line 22: It seems we need a proxy on `api`, too.
The tests of the legacy API shouldn't need to change imports.

pitercl (changes requested):

tests/unit/legacy/test_api_client.py
michalsosn: hmm, the proxy should exist, but not sure if the test should use the proxy actually. The test itself was moved into legacy package

PR #273, opened 2025-04-09
fix: functions in alpha allow positional args

michalsosn:


pitercl (changes requested):
@michalsosn Pls replace all the docs with some uniform string that explains that those methods are here just for backward compat and will be removed at some point and that they (the user) should use regular imports, not the "alpha" ones.

@normandy7 could you pls help formulating this?

pitercl (approved):

PR #277, opened 2025-04-25

michalsosn:


src/neptune_fetcher/alpha/internal/util.py
```diff
@@ -0,0 +1,81 @@
...
+    if isinstance(experiments, list):
+        return _filters.Filter.name_in(*experiments)
+    return experiments
+
+
+def resolve_attributes_filter(
+    attributes: Optional[Union[str, list[str], _filters.AttributeFilter]],
+    default_type: Optional[
+        list[
+            Literal["float", "int", "string", "bool", "datetime", "float_series", "string_set", "string_series", "file"]
```
kgodlewski, line 49: - Would it make sense to define this as a new type somewhere project level? This is not the only spot we need to typehint attr types as literals
- The name `default_type` does not make the purpose clear. `type_filter` or similar is probably closer

src/neptune_fetcher/alpha/internal/util.py
```diff
@@ -0,0 +1,81 @@
...
+            return _filters.AttributeFilter(name_eq=attributes)
+        return attributes
+    else:
+        if attributes is None:
+            return _filters.AttributeFilter(type_in=default_type)
+        if isinstance(attributes, str):
+            return _filters.AttributeFilter(name_matches_all=attributes, type_in=default_type)
+        if isinstance(attributes, list):
+            return _filters.AttributeFilter(name_eq=attributes, type_in=default_type)
+        return attributes
```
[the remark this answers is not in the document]
kgodlewski: I agree with that comment, we're duplicating important logic just to conditionally pass `type_in`

src/neptune_fetcher/alpha/internal/util.py
```diff
@@ -0,0 +1,81 @@
...
+    if isinstance(sort_by, str):
+        return _filters.Attribute(sort_by)
+    return sort_by
+
+
+def resolve_destination_path(destination: Optional[str]) -> pathlib.Path:
+    if destination is None:
+        return pathlib.Path.cwd()
+    else:
+        return pathlib.Path(destination).resolve()
```
kgodlewski, line 81: Should we error out here if the path is not a directory? We're checking for write access later, so maybe adding an `is_dir()` check wouldn't hurt, to error out early on, before we launch the download threads

michalsosn: we are creating the directory, if it does not exist
unless you mean checking if it exists, but is not a directory

kgodlewski (approved):

PR #306, opened 2025-05-22
fix: assume that series with searchAfter=null are still incomplete

michalsosn:
## Summary by Sourcery

Fix pagination logic to include series with missing `searchAfter` cursors as incomplete and correctly propagate `null` values in subsequent fetch requests.

Bug Fixes:
- Include series entries with `searchAfter = null` in new page parameters.
- Set `searchAfter` to `null` in the request payload when no previous cursor exists.

### DIFF
--- a/src/neptune_fetcher/alpha/internal/retrieval/series.py
+++ b/src/neptune_fetcher/alpha/internal/retrieval/series.py
@@ -130,23 +130,30 @@ def _make_new_series_page_params(
                 del request["searchAfter"]
         return params
 
-    request_id_to_search_after = {
-        series.requestId: series.searchAfter
-        for series in data.series
-        if series.searchAfter is not None and series.searchAfter.token and not series.searchAfter.finished
-    }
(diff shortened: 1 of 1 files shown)


src/neptune_fetcher/alpha/internal/retrieval/series.py
```diff
@@ -130,23 +130,30 @@ def _make_new_series_page_params(
...
-            search_after = request_id_to_search_after[request_id]
-            request["searchAfter"] = {
-                "finished": search_after.finished,
-                "token": search_after.token,
-            }
+        if request_id in request_id_to_search_after_token:
+            search_after_token = request_id_to_search_after_token[request_id]
+            if search_after_token is None:
+                if "searchAfter" in request:
+                    del request["searchAfter"]
```
kgodlewski, line 151:
```suggestion
            if search_after_token is None:
                request.pop("searchAfter", None)
```

pitercl (approved):

PR #311, opened 2025-05-26
feat: set user agent in http client

michalsosn:


src/neptune_fetcher/internal/client.py
```diff
@@ -55,9 +55,9 @@ def get_client(context: Context, proxies: Optional[Dict[str, str]] = None) -> Au
             return client
 
         credentials = Credentials.from_api_key(api_key=context.api_token)
-        config, token_urls = get_config_and_token_urls(credentials=credentials, proxies=proxies)
+        config, token_urls = get_config_and_token_urls(credentials=credentials, proxies=proxies, api_version="v1")
```
pitercl, line 58: Unfortunately I think we need to expose this as param of `get_client`. `internal` will soon be shared between `alpha` and `v1`.

Also, I'd like to have a distinction between `alpha` and `v1`, so for now it should be set to `alpha`.

tests/e2e/conftest.py
```diff
@@ -35,7 +35,7 @@ def client() -> AuthenticatedClient:
     credentials = Credentials.from_api_key(api_key=api_token)
     config, token_urls = get_config_and_token_urls(credentials=credentials, proxies=None)
```
pitercl, line 36: Let's set `api_version="test"` here, too, for consistency.

pitercl (changes requested):

PR #313, opened 2025-05-26
Handle `UnableToParseResponse`

kgodlewski:
## Summary by Sourcery

Prevent retrying on response parsing failures by catching `UnableToParseResponse` and raising an immediate error with proper status and content, and add a unit test to confirm this behavior.

Bug Fixes:
- Catch `UnableToParseResponse` in `backoff_retry` and raise a user-facing exception without retrying.

Enhancements:
- Update `_raise_for_response` signature to indicate it never returns.

Tests:
- Add unit test to verify no retry occurs when response parsing fails.

### DIFF
--- a/src/neptune_fetcher/internal/retrieval/util.py
+++ b/src/neptune_fetcher/internal/retrieval/util.py
@@ -30,7 +30,10 @@ from typing import (
 
 import httpx
 from neptune_api import AuthenticatedClient
-from neptune_api.errors import ApiKeyRejectedError
+from neptune_api.errors import (
+    ApiKeyRejectedError,
+    UnableToParseResponse,
+)
 from neptune_api.types import Response
 
 from neptune_fetcher import exceptions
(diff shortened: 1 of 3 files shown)


src/neptune_fetcher/internal/retrieval/util.py
```diff
@@ -109,6 +113,8 @@ def backoff_retry(
                 "Check your network connection or increase the timeout by setting the "
                 "NEPTUNE_HTTP_REQUEST_TIMEOUT_SECONDS environment variable (default: 60 seconds)."
             )
+        except UnableToParseResponse as e:
+            _raise_for_response(e.response.status_code, e.response.content)
```
michalsosn, line 117: it breaks the retry mechanism...
when is UnableToParseResponse thrown? is it definitely na unrecoverable error?

src/neptune_fetcher/internal/retrieval/util.py
```diff
@@ -109,6 +113,8 @@ def backoff_retry(
                 "Check your network connection or increase the timeout by setting the "
```
michalsosn, line 113: btw the
`raise NeptuneInvalidCredentialsError from e` above
should be 
`raise NeptuneInvalidCredentialsError() from e`

src/neptune_fetcher/internal/retrieval/util.py
```diff
@@ -142,7 +148,7 @@ def backoff_retry(
         raise error
 
 
-def _raise_for_response(status_code: int, content: bytes) -> Exception:
+def _raise_for_response(status_code: int, content: bytes) -> NoReturn:
```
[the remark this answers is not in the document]
kgodlewski: Mypy did complain at some point, this is why I added NoReturn

src/neptune_fetcher/internal/retrieval/util.py
kgodlewski: `e.response` is `httpx.Response`, not the `Response` object from `neptune_api` so we cannot assign it like you suggested. `UnableToParseResponse` is raised anytime `neptune_api` is unable to parse JSON, regardless of the status code. I decided to handle this as an unrecoverable error because there's no point in retrying a request that returns a broken response.

michalsosn (approved):

PR #346, opened 2025-06-25
feat: series retrieval fetches all types of series

michalsosn:
towards PY-143

## Summary by Sourcery

Unify series value handling by replacing the string‐only representation with a generic SeriesValue type and add support for parsing file and histogram value objects.

New Features:
- Support parsing file and histogram series values in addition to strings.

Enhancements:
- Replace StringSeriesValue with a generic SeriesValue NamedTuple carrying Any value payload.
- Introduce an extraction helper to convert ProtoPointValueDTO into the appropriate Python object based on its field.
- Update type hints and references across retrieval, composition, and output modules to use the new SeriesValue.

Tests:
- Update existing unit and end-to-end tests to reference SeriesValue and validate series retrieval with various value types.

### DIFF
--- a/src/neptune_fetcher/internal/filters.py
+++ b/src/neptune_fetcher/internal/filters.py
@@ -123,7 +123,7 @@ class _AttributeFilter(_BaseAttributeFilter):
     """
 
     name_eq: Union[str, list[str], None] = None
-    type_in: Sequence[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(types.ALL_TYPES))  # type: ignore
+    type_in: Sequence[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(types.ALL_TYPES))
     must_match_any: Optional[list[_AttributeNameFilter]] = None
     aggregations: Sequence[AGGREGATION_LITERAL] = field(default_factory=lambda: ["last"])
 
@@ -139,8 +139,8 @@ class _AttributeFilter(_BaseAttributeFilter):
                 _validate_string_list(item.must_match_regexes, "must_match_regexes")
                 _validate_string_list(item.must_not_match_regexes, "must_not_match_regexes")
 
-        _validate_list_of_allowed_values(self.type_in, types.ALL_TYPES, "type_in")  # type: ignore
-        _validate_list_of_allowed_values(self.aggregations, types.ALL_AGGREGATIONS, "aggregations")  # type: ignore
+        _validate_list_of_allowed_values(self.type_in, types.ALL_TYPES, "type_in")
+        _validate_list_of_allowed_values(self.aggregations, types.ALL_AGGREGATIONS, "aggregations")
 
     def transform(
         self, map_attribute_filter: Callable[["_AttributeFilter"], "_AttributeFilter"]
--- a/src/neptune_fetcher/internal/filters.py
+++ b/src/neptune_fetcher/internal/filters.py
@@ -25,6 +25,7 @@ from typing import (
     Iterable,
     Literal,
     Optional,
+    Sequence,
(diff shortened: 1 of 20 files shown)


src/neptune_fetcher/internal/filters.py
```diff
@@ -112,9 +123,9 @@ class _AttributeFilter(_BaseAttributeFilter):
     """
 
     name_eq: Union[str, list[str], None] = None
-    type_in: list[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(types.ALL_TYPES))  # type: ignore
+    type_in: Sequence[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(types.ALL_TYPES))
```
gabrys, line 126: On a related note: I wanted to audit the code for `list[...]` and start replacing them with `Sequence[...]`. The issue I found is sometimes we're only allowing lists by checking `instanceof(list)` so this would require more checking...

src/neptune_fetcher/internal/output_format.py
```diff
@@ -270,7 +256,7 @@ def generate_categorized_rows() -> Generator[Tuple, None, None]:
 
 
 def create_series_dataframe(
-    series_data: dict[identifiers.RunAttributeDefinition, list[series.StringSeriesValue]],
+    series_data: dict[identifiers.RunAttributeDefinition, list[series.SeriesValue]],
```
gabrys, line 259: similarly to list->Sequence, dict could be a Mapping here (pending verification it's actually working)

src/neptune_fetcher/internal/util.py
```diff
@@ -35,13 +37,13 @@ def _validate_string_list(value: Optional[list[str]], field_name: str) -> None:
             raise ValueError(f"{field_name} must be a list of strings")
 
 
-def _validate_list_of_allowed_values(value: list[str], allowed_values: set[str], field_name: str) -> None:
+def _validate_list_of_allowed_values(value: Sequence[str], allowed_values: Collection[str], field_name: str) -> None:
     """Validate that a value is a list containing only allowed values."""
     if not isinstance(value, list) or not all(isinstance(v, str) and v in allowed_values for v in value):
```
gabrys, line 42: Here, we're checking against a list specifically, but promise to accept any Sequence. Let's maybe revert to typing as list for now.

gabrys (approved):
I think, change `typing.Sequence` back to `list` and ship it :shipit: :+1:

src/neptune_fetcher/internal/util.py
michalsosn:
```
src/neptune_fetcher/alpha/filters.py:110: error: Argument 1 to "_validate_list_of_allowed_values" has incompatible type "list[Literal['float', 'int', 'string', 'bool', 'datetime', 'float_series', 'string_set', 'string_series', 'file']]"; expected "list[str]"  [arg-type]
src/neptune_fetcher/alpha/filters.py:110: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/neptune_fetcher/alpha/filters.py:110: note: Consider using "Sequence" instead, which is covariant
src/neptune_fetcher/alpha/filters.py:111: error: Argument 1 to "_validate_list_of_allowed_values" has incompatible type "list[Literal['last', 'min', 'max', 'average', 'variance']]"; expected "list[str]"  [arg-type]
src/neptune_fetcher/alpha/filters.py:111: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/neptune_fetcher/alpha/filters.py:111: note: Consider using "Sequence" instead, which is covariant
```

PR #353, opened 2025-06-26
PY-122 Improve the API of Filter and AttributeFilter in v1

gabrys:
`Filter`

* `matches_all` & `matches_none` -> `matches` (with ERS support)
* `name_in` & `name_eq` -> `name()` (with ERS support if single string, and direct match alternative if list of strings)

`AttributeFilter`

* `name_matches_all` & `name_matches_none` -> remove
* `name_eq` -> `name` (with ERS support if single string, and direct match alternative if list of strings)
* `type_in` -> `type`

## Summary by Sourcery

Improve the filters API in v1 by unifying name-based criteria under a single `name` parameter, streamlining regex handling, and enforcing non-empty input lists.

New Features:
- Replace `name_eq`, `name_matches_all`, and `name_matches_none` with a unified `name` parameter in `AttributeFilter` and `Filter`.
- Add a new `Filter.name` method to match experiment names via extended regex or exact lists.
- Introduce strict validation to raise errors on empty lists for experiments, runs, and attributes filters.

Enhancements:
- Leverage core `_filters._Filter` operators for AND, OR, and NOT logic and remove custom static methods.
- Extend `build_extended_regex_attribute_filter` to accept optional `type_in` and `aggregations` parameters.
- Simplify internal filter resolution by constructing attribute and experiment filters directly without intermediate wrappers.

Tests:
- Update end-to-end and unit tests to replace deprecated filter methods (`name_in`, `matches_all`, `matches_none`) with the new API.
- Adjust test expectations for empty-list error handling in experiments, runs, and attributes filtering.

### DIFF
--- a/src/neptune_fetcher/v1/filters.py
+++ b/src/neptune_fetcher/v1/filters.py
@@ -92,11 +92,11 @@ class AttributeFilter(BaseAttributeFilter):
         name (str|list[str], optional):
             if str given: an extended regular expression to match attribute names.
             if list[str] given: a list of attribute names to match exactly.
-        type_in (list[Literal["float", "int", "string", "bool", "datetime", "float_series", "string_set",
-        "string_series", "file"]]):
+        type (list[Literal["float", "int", "string", "bool", "datetime", "float_series", "string_set",
+        "string_series", "file"]], optional):
             A list of allowed attribute types. Defaults to all available types.
             For a reference, see: https://docs.neptune.ai/attribute_types
-        aggregations (list[Literal["last", "min", "max", "average", "variance"]]): List of
+        aggregations (list[Literal["last", "min", "max", "average", "variance"]], optional, deprecated): List of
             aggregation functions to apply when fetching metrics of type FloatSeries or StringSeries.
             Defaults to ["last"].
 
@@ -108,7 +108,7 @@ class AttributeFilter(BaseAttributeFilter):
 
 
     loss_avg_and_var = AttributeFilter(
-        type_in=["float_series"],
+        type=["float_series"],
         name="loss$",
         aggregations=["average", "variance"],
     )
@@ -118,25 +118,25 @@ class AttributeFilter(BaseAttributeFilter):
     """
 
     name: Union[str, list[str], None] = None
-    type_in: list[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(KNOWN_TYPES))  # type: ignore
+    type: list[ATTRIBUTE_LITERAL] = field(default_factory=lambda: list(KNOWN_TYPES))  # type: ignore
     aggregations: list[AGGREGATION_LITERAL] = field(default_factory=lambda: ["last"])
 
     def __post_init__(self) -> None:
         _validate_string_or_string_list(self.name, "name")
-        _validate_list_of_allowed_values(self.type_in, KNOWN_TYPES, "type_in")
+        _validate_list_of_allowed_values(self.type, KNOWN_TYPES, "type")
         _validate_list_of_allowed_values(self.aggregations, ALL_AGGREGATIONS, "aggregations")
 
     def _to_internal(self) -> _filters._AttributeFilter:
         if isinstance(self.name, str):
             return _pattern.build_extended_regex_attribute_filter(
                 self.name,
-                type_in=self.type_in,
+                type_in=self.type,
                 aggregations=self.aggregations,
             )
 
         if self.name is None:
             return _filters._AttributeFilter(
-                type_in=self.type_in,
+                type_in=self.type,
                 aggregations=self.aggregations,
             )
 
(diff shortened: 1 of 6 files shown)


src/neptune_fetcher/v1/filters.py
```diff
@@ -297,104 +304,49 @@ def matches(attribute: Union[str, Attribute], pattern: str) -> "Filter":
...
+            _filters._AttributeValuePredicate(operator="CONTAINS", attribute=attribute._to_internal(), value=value)
+            for value in values
+        ]
+
+        return Filter(_filters._Filter.all(internal_filters))
 
     @staticmethod
-    def contains_none(attribute: Union[str, Attribute], value: Union[str, list[str]]) -> "Filter":
+    def contains_none(attribute: Union[str, Attribute], values: list[str]) -> "Filter":
         if isinstance(attribute, str):
```
michalsosn, line 323: str is no longer accepted

src/neptune_fetcher/v1/_internal.py
```diff
@@ -63,42 +70,42 @@ def resolve_attributes_filter(
...
-            # In v1, passing attributes=[] gives us un-filtered results
-            # In v1, we're going to return no results or raise an error
-            return _filters._AttributeFilter()
+            raise ValueError(
+                "Invalid type for `attributes` filter. Expected str, non-empty list of str, or AttributeFilter object, "
+                "but got empty list."
+            )
         if isinstance(attributes, list):
-            return filters.AttributeFilter(name_eq=attributes)._to_internal()
+            return _filters._AttributeFilter(name_eq=attributes)
```
michalsosn, line 78: hmm, in most cases you are building a v1 filter and calling _to_internal, but in this you are building an internal filter directly

gabrys: I can see the point, but `_pattern.build_extended_regex_attribute_filter` already only returns an internal `_Filter`. And I'm not sure we can express all the filters we're creating here as non-internal ones.

src/neptune_fetcher/v1/filters.py
```diff
@@ -297,104 +304,49 @@ def matches(attribute: Union[str, Attribute], pattern: str) -> "Filter":
...
-        if isinstance(regex, str):
-            return Filter(
-                _filters._AttributeValuePredicate(operator="MATCHES", attribute=attribute._to_internal(), value=regex)
-            )
-        else:
-            filters = [Filter.matches_all(attribute, r) for r in regex]
-            return Filter.all(*filters)
 
-    @staticmethod
-    def matches_none(attribute: Union[str, Attribute], regex: Union[str, list[str]]) -> "Filter":
```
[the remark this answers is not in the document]
gabrys: I'm leaning slightly towards reverting to `Union[str, list[str]]` although contains_all/_none accepting a single string doesn't resonate with me...

src/neptune_fetcher/internal/pattern.py
```diff
@@ -92,9 +95,17 @@ def build_extended_regex_filter(attribute: _Attribute, pattern: str) -> _Filter:
...
+def build_extended_regex_attribute_filter(
+    pattern: str,
+    type_in: Optional[list[ATTRIBUTE_LITERAL]] = None,
+    aggregations: Optional[list[AGGREGATION_LITERAL]] = None,
+) -> _AttributeFilter:
     parsed = parse_extended_regex(pattern)
 
+    # properly set defaults:
+    type_in = ALL_TYPES if type_in is None else type_in
+    aggregations = ["last"] if aggregations is None else aggregations
```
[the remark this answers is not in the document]
gabrys: on the other hand, the ALL_TYPES is defined in internal, not in alpha/v1/etc... Maybe that can move to version specific module... :thinking:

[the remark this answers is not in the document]
michalsosn: yeah
ALL_TYPES as a list of all types supported internally can stay (it may be used somewhere)
and then v1 and alpha should have their own copy ALL_TYPES used as the default in filters and maybe validation...

As for how useful it is, I think I'd want to add "histogram_series" to the v1's ALL_TYPES but not support in alpha (?)

src/neptune_fetcher/v1/filters.py
```diff
@@ -88,47 +83,52 @@ class AttributeFilter(BaseAttributeFilter):
...
+                "Invalid type for `name` attribute. Expected str, non-empty list of str, or None, but got empty list."
+            )
 
-        return _filters._AttributeFilter(
-            name_eq=self.name_eq,
-            type_in=self.type_in,
-            must_match_any=must_match_any,
-            aggregations=self.aggregations,
+        if isinstance(self.name, list):
+            return _filters._AttributeFilter.name_eq(
```
pitercl, line 123: This is a bug: `name_eq` is not callable.

michalsosn (approved):

PR #364, opened 2025-07-02
fix: ignore inference errors when there are no runs + return more details in inference exception

michalsosn:
fixes PY-153

## Summary by Sourcery

Refine attribute type inference errors to provide clearer, context-specific messages and update tests accordingly

Enhancements:
- Differentiate error messages for missing attributes vs. conflicting types
- Pass conflicting types into AttributeTypeInferenceError for more detailed feedback

Tests:
- Rename and update tests to capture exceptions and assert the new error messages

### DIFF
--- a/src/neptune_fetcher/internal/composition/download_files.py
+++ b/src/neptune_fetcher/internal/composition/download_files.py
@@ -80,8 +80,7 @@ def download_files(
                 {},
                 index_column_name="experiment" if container_type == search.ContainerType.EXPERIMENT else "run",
             )
-        inference_result.raise_if_incomplete()
-        filter_ = inference_result.result
+        filter_ = inference_result.get_result_or_raise()
 
         sys_id_label_mapping: dict[identifiers.SysId, str] = {}
 


src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -84,34 +170,32 @@ def infer_attribute_types_in_sort_by(
...
         container_type=container_type,
+        inference_state=state,
     )
-    attributes = _filter_untyped(attributes)
-    if attributes:
-        raise AttributeTypeInferenceError(attribute_names=[a.name for a in attributes])
+    return state
 
 
 def _infer_attribute_types_from_attribute(
```
pitercl, line 194: `_infer_attribute_types_locally`, maybe?

src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -122,64 +206,71 @@ def _infer_attribute_types_from_attribute(
...
+                ),
+                lambda: concurrency.return_value(sys_ids_page.items),
+            ],
+        ),
     )
 
-    attribute_definition_pages: Generator[
-        util.Page[identifiers.AttributeDefinition], None, None
-    ] = concurrency.gather_results(output)
+    results: Generator[util.Page[identifiers.AttributeDefinition], None, None] = concurrency.gather_results(output)
```
pitercl, line 250: This type hint could express that sys_ids are also returned here.

src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -122,64 +206,71 @@ def _infer_attribute_types_from_attribute(
...
-        raise RuntimeError(f"Unexpected filter type: {type(experiment_filter)}")
+            types = attribute_name_to_definition[attribute.name]
+            if len(types) == 1:
+                state.set_success(
+                    inferred_type=next(iter(types)),  # type: ignore
+                    success_details="Inferred from neptune api",
+                )
+            elif len(types) > 1:
+                state.set_error(
+                    error=f"Neptune found the attribute name in multiple runs "
```
pitercl, line 272:
```suggestion
                    error=f"Neptune found the attribute name in multiple {'runs' if container_type == ContainerType.RUN else 'experiments'} "
```

src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -42,38 +44,122 @@
...
+            for attr_state in uninferred_attributes:
+                if attr_state.error:
+                    details.append(f"{attr_state.original_attribute.name}: {attr_state.error}")
+                else:
+                    details.append(f"{attr_state.original_attribute.name}: could not find the attribute")
+
+            raise AttributeTypeInferenceError(attribute_names=attribute_names, details=details)
+
+
 def infer_attribute_types_in_filter(
```
pitercl, line 134: The API of this method (and `infer_attribute_types_in_sort_by`, too) is quite unintuitive for me, because it returns something and yet its actual result is passed via mutation of arguments.
What would you say to:
1. Copying the Filter and working on a copy internally.
2. Returning some `InferenceResult` object with an API:
    1. `raise_if_incomplete`
    2. `is_run_domain_empty`
    3. `inferred_filter` (the copy)

(and similarly for the second method)

src/neptune_fetcher/internal/composition/list_containers.py
```diff
@@ -36,19 +36,21 @@ def list_containers(
...
     ):
-        type_inference.infer_attribute_types_in_filter(
+        inference_result = type_inference.infer_attribute_types_in_filter(
             client=client,
             project_identifier=project_identifier,
             filter_=filter_,
             executor=executor,
             fetch_attribute_definitions_executor=fetch_attribute_definitions_executor,
         )
+        if inference_result is None:
```
pitercl, line 51: This cannot happen, can it? We probably want `is_run_domain_empty` here?

pitercl (changes requested):

src/neptune_fetcher/internal/composition/list_containers.py
michalsosn: hmm, copilot decided to generate a wrong line in just this one function

src/neptune_fetcher/internal/composition/type_inference.py
michalsosn: I'm a bit worried copy.deepcopy could fail for some type but seems fine

PR #384, opened 2025-07-14
Use empty attribute list by default in fetch_x_table

gabrys:
This is #368 applied on the top of the current `main` post the introduction of `neptune_query`.

## Summary by Sourcery

Enable empty attribute lists by default in fetch_experiments_table and fetch_runs_table, implement an empty-attribute filter type, adjust filter logic to skip fetching when no attributes are specified, and add tests for the new behavior

New Features:
- Support passing an empty attribute list to fetch_experiments_table and fetch_runs_table
- Default the attributes parameter to an empty list in both fetch functions

Bug Fixes:
- Remove errors when an empty list is provided for attributes and handle it gracefully

Enhancements:
- Introduce _EmptyAttributeFilter and update filter resolution, conversion, splitting, and composition to handle empty attribute filters

Tests:
- Add end-to-end tests for empty attribute list scenarios in experiments and runs table fetching

### DIFF
--- a/src/neptune_query/filters.py
+++ b/src/neptune_query/filters.py
@@ -19,6 +19,7 @@ from datetime import datetime
 from typing import (
     Iterable,
     Literal,
+    Sequence,
     Union,
 )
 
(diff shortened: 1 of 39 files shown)


pitercl (approved):

src/neptune_query/filters.py
```diff
@@ -117,7 +117,7 @@ def __post_init__(self) -> None:
         _validate_string_or_string_list(self.name, "name")
         _validate_list_of_allowed_values(self.type, KNOWN_TYPES, "type")
 
-    def _to_internal(self) -> _filters._AttributeFilter:
+    def _to_internal(self) -> Union[_filters._AttributeFilter, _filters._EmptyAttributeFilter]:
```
michalsosn, line 120: _BaseAttributeFilter could be more appropriate

PR #398, opened 2025-07-28
Remove pre-optimization in fetching attribute definitions

gabrys:
The optimization tries to avoid multi-threading for fetching attribute definitions if only one Filter is passed to `_fetch_attribute_definitions`.

However it doesn't make the code any faster and introduces unnecessary code paths.

Updated unit tests to match the simplified code.

Tested that avoiding calling `concurrency.generate_concurrently` with one item doesn't gain any measurable performance, so I'm removing the extra code from `internal/composition/attributes.py` for simplicity.

### DIFF
--- a/neptune_fetcher/src/neptune_fetcher/internal/composition/attributes.py
+++ b/neptune_fetcher/src/neptune_fetcher/internal/composition/attributes.py
@@ -128,18 +128,13 @@ def _fetch_attribute_definitions(
 
     filters_ = att_defs.split_attribute_filters(attribute_filter)
 
-    if len(filters_) == 1:
-        head = filters_[0]
-        for page in go_fetch_single(head):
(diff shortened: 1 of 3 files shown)


gabrys:
> Didn't you encounter some bug when changing this code? https://linear.app/neptuneai/issue/PY-169/investigate-infinite-thread-creation-in-af8745 ? What happened to it in the end?

In the end this was due to this missing line: https://github.com/neptune-ai/neptune-fetcher/pull/398/files#diff-531eff872625bb94d3de1adeb78104e38c043d0df438588770f253ec001283d1R56

MagickMock require extra caution when working with generators :shrug: . It came down to `next` not modifying the state of the generator and thus generating an infinite list of items to process...

pitercl (approved):

PR #401, opened 2025-07-31
Fix type inference - no more downloading all sys ids

gabrys:
1. Added a list of known sys arguments and their types, to avoid calling API to resolve these

2. Stop fetching all sys_ids to infer artibute types

3. No longer treat empty project specially

Consequences:

1. Much quicker type inference -- no longer dependent on the size of the project, no more flood of requests to fetch all sys_ids

2. At the cost of users may now be more often asked to specify the type explicitly, for example:

   * `sort_by` inference no longer takes into account the resulting experiments/runs
   * Type interference now cannot filter types between experiments and runs

## Summary by Sourcery

Improve type inference by using a static map for sys attribute types, removing empty-project special cases, and eliminating costly sys_id downloads.

Bug Fixes:
- Prevent fetching all system IDs for type inference to reduce request floods and improve performance

Enhancements:
- Infer system attributes locally using a static KNOWN_ATTRIBUTE_TYPES map to avoid API resolution
- Remove run_domain_empty flag and eliminate special-case handling of empty projects and containers
- Streamline API usage by fetching attribute definitions directly instead of generating sys_id lists
- Default unset sort_by type to string and clean up unused parameters in inference functions

Tests:
- Update tests to remove assertions on run_domain_empty and mark irrelevant tests as skipped

### DIFF
--- a/neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
+++ b/neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
@@ -17,12 +17,10 @@ from collections import defaultdict
 from concurrent.futures import Executor
 from dataclasses import dataclass
 from typing import (
-    Generator,
     Generic,
     Iterable,
     Optional,
     TypeVar,
-    Union,
 )
 
 from neptune_api.client import AuthenticatedClient
@@ -32,12 +30,7 @@ from .. import (
     filters,
     identifiers,
 )
-from ..composition import attribute_components as _components
-from ..composition import concurrency
-from ..retrieval import (
-    search,
-    util,
-)
+from ..retrieval import search
 from ..retrieval.attribute_types import (
     ATTRIBUTE_LITERAL,
     FILE_SERIES_AGGREGATIONS,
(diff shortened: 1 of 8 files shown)


neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -245,58 +281,34 @@ def _infer_attribute_types_from_api(
...
-                    attribute_filter=attribute_filter_by_name,
-                    executor=executor,
-                    fetch_attribute_definitions_executor=fetch_attribute_definitions_executor,
-                    sys_ids=sys_ids_page.items,
-                    downstream=lambda _, definitions: concurrency.return_value(definitions),
-                ),
-                lambda: concurrency.return_value(sys_ids_page.items),
-            ],
-        ),
+    output = fetch_attribute_definitions(
```
michalsosn, line 284: you've said you are aware of this, but still note that by skipping the fetch_sys_ids call, you'll read attributes from both runs and experiments, even when container_type is EXPERIMENT. 
Calling fetch_sys_ids is needed to filter out the runs that represent only the EXPERIMENTs

neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
```diff
@@ -245,58 +281,34 @@ def _infer_attribute_types_from_api(
...
-                    success_details="Inferred from neptune api",
-                )
-            elif len(types) > 1:
-                container_name = "runs" if container_type == ContainerType.RUN else "experiments"
-                state.set_error(
-                    error=f"Neptune found the attribute name in multiple {container_name} "
-                    f"with conflicting types: {', '.join(types)}"
-                )
-
-    inference_state.run_domain_empty = len(sys_ids) == 0
```
michalsosn, line 302: this is also something that you're aware of, but still I'll note that this run_domain_empty was introduced in response to https://linear.app/neptuneai/issue/PY-153/fetch-experiments-table-blows-up-on-attribute-inference-produces-a 
i.e. to detect a situation where the inference failed not because an attribute was not found, but because there were no runs to read attributes from in the first place and return an empty result in such case

neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
gabrys: Yes. This feature (filter attributes by runs/experiment for the purpose of type inference) is definitely not worth the extra complexity and performance hit coming from fetching all sys_ids.

neptune_fetcher/src/neptune_fetcher/internal/composition/type_inference.py
gabrys: Yup, the completely empty project is a big edge case, not worth having this much complexity for handling this.

gabrys:
This is superseded by #406

PR #442, opened 2026-01-15
PY-317 Bundle neptune-api in neptune-fetcher

gabrys:

## Summary by Sourcery

Vendor the neptune-api client and protobuf definitions into the neptune_fetcher package and update internal retrieval, auth, and tests to use the bundled API with stricter response validation.

New Features:
- Embed a generated neptune_api client, models, and protobuf definitions under neptune_fetcher.generated.neptune_api to remove the external neptune-api runtime dependency.

Bug Fixes:
- Validate types and presence of parsed configuration and token URL responses before use, raising clear runtime errors when data is missing or malformed.
- Map rejected API keys to a dedicated invalid-credentials error without retrying, while preserving retry behavior for retriable response parsing failures.
- Fail fast when no API token is provided in the client context instead of proceeding with invalid credentials.
- Guard retrieval paths against missing parsed data when fetching attribute definitions and signed file URLs.

Enhancements:
- Annotate protobuf deserialization helpers with explicit DTO variables to clarify return types and improve static type checking.
- Switch internal imports to the bundled neptune_api client types to keep retrieval utilities self-contained.

Build:
- Replace the direct neptune-api dependency with its transitive dependencies in pyproject.toml and register the bundled neptune_fetcher.generated.neptune_api package for mypy.
- Exclude the generated neptune_api code from pre-commit checks and add attrs as an additional mypy dependency.

Tests:
- Update unit and package tests to import API clients and protobuf models from neptune_fetcher.generated.neptune_api and adjust mocks to the new module paths.


src/neptune_fetcher/internal/retrieval/files.py
```diff
@@ -59,7 +59,10 @@ def fetch_signed_urls(
 
     response = retry.handle_errors_default(signed_url.sync_detailed)(client=client, body=body)
 
-    data: CreateSignedUrlsResponse = response.parsed
+    data = response.parsed
```
pitercl:
```suggestion
    data: Optional[CreateSignedUrlsResponse] = response.parsed
```

pitercl (approved):

